PR #112, opened 2019-07-16
revert PR #99

Kurtil:
Next commit 0baf81a9eebba30c6b86949ba44d09670bc600ad makes this PR invalid and it needs to be reverted .

PR #113, opened 2019-07-16
Revert PR #98

Kurtil:
Next commit 0baf81a9eebba30c6b86949ba44d09670bc600ad makes this PR invalid and it needs to be reverted .


Amoki:
Same as https://github.com/xeokit/xeokit-sdk/pull/112

PR #170, opened 2019-09-30
Objects not properly deregistered from scene when destroyed

Kurtil:


Kurtil:
It seems that other kind of entities like Mesh already use the API I fixed in Scene.js (missing parameter) :
https://github.com/xeokit/xeokit-sdk/blob/master/src/viewer/scene/mesh/Mesh.js#L1709-L1739

PR #180, opened 2019-10-15
setViewpoint not immediate

Kurtil:
instead of setting camera properties, setViewpoint method of BCFViewpointsPlugin could use cameraFlight.flyTo to make the process smoother.

Warning : i did not test cameraFlight presence before calling flyTo (line 428). If cameraFlight can be absent on viewer, this should be updated.


Kurtil:
Also this change is maybe to different from the original behaviour. Indeed, it is not possible to immediately set camera like before. A config should be more appropriate.

PR #270, opened 2020-03-03
Added support for BIMServer 1.5.180

nissimpradhan:
2 major changes which makes this support possible
	- 1) Addition of new bit in the binary format
		- Compared BIMServerPerformanceGeometryLoader.js in Xeokit
		- with GeometryLoader.js in BIMVie.ws( They have the same function)
  		- and found out that the new BIMServer protocol ( used in BIMvie.ws
that comes with BIMServer 1.5.180 ) for downloading an IFC
		- invoked via ServiceInterface.download
		- has added a new Long bit for each unit of binary to be decoded in
it's binary format after the first Long ( check changed code )
		- this new long was given the name type, i guess it means the type of
message.
		- this new Long bit was the reason Xeokit was failing to parse the
rest of the data.
	  - 2) GeometryLoader.js in BIMVie.ws supports a new geometry type 3,
the code for this new geometry type was ported to Xeokit, but
incompatible lines were commented. So all in all geometry type 3 is
still not supported

### DIFF
--- a/src/plugins/BIMServerLoaderPlugin/lib/BIMServerPerformanceGeometryLoader.js
+++ b/src/plugins/BIMServerLoaderPlugin/lib/BIMServerPerformanceGeometryLoader.js
@@ -35,7 +35,9 @@ function BIMServerPerformanceGeometryLoader(bimServerClient, bimServerClientMode
     function processMessage(stream) {
         var messageType = stream.readByte();
         if (messageType === 0) {
-            readStart(stream);
+        	if (!readStart(stream)) {
+				return false;
+			}
         } else if (messageType === 6) {
             readEnd(stream);
         } else {
@@ -51,9 +53,14 @@ function BIMServerPerformanceGeometryLoader(bimServerClient, bimServerClientMode
         while (data != null) {
             stream = new DataInputStreamReader(data);
             var topicId = stream.readLong();
-            while (processMessage(stream)) {
+            var type = stream.readLong();
+            if (type == 0) {
+            	while (processMessage(stream)) {
 
-            }
+                }
+			} else if (type == 1) {
+				// End of stream
+			}
             data = todo.shift();
         }
     };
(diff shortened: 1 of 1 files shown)


xeolabs:
@nissimpradhan - just a quick bit of bureaucracy - could I please get you to e-sign the Contributor's Agreement and mail it back to me? Then I'll merge, send over a license etc. https://github.com/xeokit/xeokit.io/wiki/Contributor-License-Agreement

nissimpradhan:
@xeolabs yes of course, thanks! :) just sent it to lindsay.kay@xeolabs.com

There is another issue in Xeokit that I've found fixed. I'll commit and push it tommorow.
The issue is about not being able to load the TreeViewer when loading models from the BIM Server.

nissimpradhan:
@xeolabs Hi again, I just came to think about it now but, It might cause compatibility issue for you as the previous binary before bimserver 1.5.180 this might not be parseable anymore as it includes  code to read more than what's there( for older version). 

This is not neccessary for me as i use 1.5.180 but let me know if you want me to help you fix it.

xeolabs:
Hi @nissimpradhan, which version of BIMServer are you having success with at the moment, if any?

Just checking because I'm working on getting BIMServer examples working again, however from BIMServer 1.5.182 I'm always getting a ````NullPointerException```` when attempting to load the IFC2x3 Duplex model using ````BIMServerLoaderPlugin````.


````
18:59:10 ERROR [pool-1-thread-1]:                                                                                  (BimServer.java:240) 
java.lang.NullPointerException: null
	at org.bimserver.serializers.binarygeometry.BinaryGeometryMessagingStreamingSerializer.writeGeometryData(BinaryGeometryMessagingStreamingSerializer.java:974)
	at org.bimserver.serializers.binarygeometry.BinaryGeometryMessagingStreamingSerializer.writeData(BinaryGeometryMessagingStreamingSerializer.java:771)
	at org.bimserver.serializers.binarygeometry.BinaryGeometryMessagingStreamingSerializer.writeMessage(BinaryGeometryMessagingStreamingSerializer.java:275)
	at org.bimserver.servlets.Streamer$1.run(Streamer.java:166)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
````

nissimpradhan:
Hi @xeolabs, it is successfully working with 1.5.180, I haven't tested with 1.5.182 yet.

Also try to open an ifc4 file, maybe it works?

PR #273, opened 2020-03-05
Fixes: Tree Plugin not working when loading models from the BIMServer

nissimpradhan:


xeolabs:
@nissimpradhan good catch - just one change - could you instead use this method to register the model, then the Scene's cache of model IDs gets invalidated: https://github.com/xeokit/xeokit-sdk/blob/master/src/viewer/scene/scene/Scene.js#L915

nissimpradhan:
Please note that model still doesn't auto add itself to the scene it seems. So we need to  set off the autoAddModels and then add it after the loading is done.

```
                                                const treeView = new TreeViewPlugin(viewer, {
						    containerElement: document.getElementById("treeViewContainer"),
						    autoAddModels: false,  // <<---------------- Don't auto-add models
						    autoExpandDepth: 2,
						    hierarchy: "storeys"
						});

	    			                model.on("loaded", function() {
	    			                    //viewer.cameraFlight.jumpTo(model);
	    			                    viewer.cameraFlight.flyTo(model);
	    			                    treeView.addModel(model.id);
	    			                });
```

nissimpradhan:
@xeolabs Good morning. I'm not sure if you received my email with the signed contributor's agreement form. Did you?

xeolabs:
@nissimpradhan got it thanks. Will get the license to you today/tomorrow as well.

PR #281, opened 2020-03-20
Remove opacity change in BCF viewpoint reset

Amoki:
The removed line makes all the windows to be opaque. My solution is not the best (the opacity is not reset), but the current implementation seems odd and doesn't provide the expected behavior. And the BCF can't change the opacity. It seems to be a good trade of.


Amoki:
fixed by https://github.com/xeokit/xeokit-sdk/commit/30004f3dfe822a93f4841389a94e422c6854ffaa

PR #295, opened 2020-05-04
Fix: Fix DistanceMeasurement plugin length property

juho-ylikyla:
The DistanceMeasurementPlugin has class `DistanceMeasurement ` with the property `length` which is accessed via a getter.

The this._length was never initialized and calling the plugins length getter would always return NaN

With this PR, the this._length is set to the actual length value.


xeolabs:
Thanks @juho-ylikyla !

PR #358, opened 2020-06-22
add rotateRighClick cameraControl configuration

Kurtil:
Add the ability to disable rotate when mouse move after right-clicking.

### DIFF
--- a/src/viewer/scene/CameraControl/CameraControl.js
+++ b/src/viewer/scene/CameraControl/CameraControl.js
@@ -451,6 +451,7 @@ class CameraControl extends Component {
             followPointer: false,
             doublePickFlyTo: true,
             panRightClick: true,
+            rotateRightClick: true,
             showPivot: false,
             pointerEnabled: true,
             constrainVertical: false,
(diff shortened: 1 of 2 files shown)


Amoki:
@xeolabs Do you have something to say about this PR?

src/viewer/scene/CameraControl/CameraControl.js
```diff
@@ -921,6 +923,28 @@ class CameraControl extends Component {
...
 
+    /**
+     * Sets whether mouse-moving after right-click rotates the {@link Camera}.
+     *
+     * Default is ````true````.
+     *
+     * @param {Boolean} value Set ````false```` to disable rotate on mouse-move after right-click.
+     */
+    set rotateRightClick(value) {
+        this._configs.rotateRightClick = value !== false;
```
xeolabs, line 934: I think this should be false by default, so that left-drag-rotate is able to be default behaviour

Kurtil: @xeolabs , the current behaviour is `true`. If we put `false` this will lead to minor breaking change... But if you think this is better I can update it.

Kurtil:
I close it for inactivity.

PR #415, opened 2020-08-14
Improve CameraControl touch dolly/pan

Amoki:
But when doing a zoom, a pick is done at each tick, and the pick costs ~60% of the time (cf. flame graph).
Capture d’écran de 2020-08-14 16-54-15

Checking the distance only once when the two fingers are touching the screen and keeping it until a finger leaves the screen seems a good UX and is more performant.

cameraUpdater doesn't pick anything itself anymore, each handler must explicitly pick before a dolly.

Touch events were triggered for every pixel, implying 150 updates of 1 pixel instead of 1 update of 150px. This led to very slow dolly.

I retored my `viewer.cameraControl.touchDollyRate = 0.2;` to avoid too fast dolly.

Pan and dolly can also be done simultaneously: it's way more smooth and a better UX overall.


xeolabs:
@Amoki thanks very much!

PR #497, opened 2020-12-10
simplify BCF plugin

Amoki:
If `objectId !== originalSystemId`, no ifc_guid was defined, so no other BCF tools could use the viewpoint.

Now, `ifc_guid` is always defined and `authoring_tool_id` is defined if it provides more information.

### DIFF
--- a/src/plugins/BCFViewpointsPlugin/BCFViewpointsPlugin.js
+++ b/src/plugins/BCFViewpointsPlugin/BCFViewpointsPlugin.js
@@ -398,13 +398,11 @@ class BCFViewpointsPlugin extends Plugin {
         const xrayedObjectIds = new Set(scene.xrayedObjectIds);
         const colorizedObjectIds = new Set(scene.colorizedObjectIds);
 
-        let authoringComponentsCreated = false;
-        const originalSystemIdMap = {};
         const originalSystemColoringMap = {};
 
-        const authoringSystemColoringMap = Object.values(scene.objects)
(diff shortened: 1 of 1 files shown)


xeolabs:
Ah OK. I misunderstood the BCF spec and thought that the ````ifc_guid```` was optional for a component when ````authoring_tool_id```` was supplied. This makes sense.

However, when saving such a viewpoint, there will be many components that have unique ````authoring_tool_id```` values, but the same ````ifc_guid```` value. So, when loading such a viewpoint, I hope that doesn't mean a lot of redundant, repeated setting of state on objects referenced by the repeated ````ifc_guid```` values.  

This was my rationale for only providing ````authoring_tool_id```` on those components, without ````ifc_guid````, so that we don't have to have components with different ````authoring_tool_id```` but the same ````ifc_guid````.

I was also assuming that a BCF viewpoint with multiple instances of a model would only be loadable into xeokit anyway, where we can't expect another viewer to understand the globalized IDs in the multiple model instances, and multiple components with the same ````ifc_guid```` would not make sense to other viewers.

However, I'm willing to merge this PR if you think that these issues aren't a problem for us.

Amoki:
If we compare models on xeokit and create BCF, it's mostly to tell one version of a model has a problem. We can suppose only one model will be opened in an editor to be corrected, so keeping the original `ifc_guid` makes sense here.

Duplicated `ifc_guid` shouldn't be a problem.

xeolabs:
@Amoki I've released this in 1.5.22

PR #522, opened 2021-01-07
Remove default cursor style.

Kurtil:
If "default" is added to the cursor style, the cursor style from parent DOM Elements won't be able to change the canvas cursor style.

If null (or "") is added instead, it removes the cursor style entry from the style attribute of the canvas element, letting cursor style cascading from parent elements.


Kurtil:
Use case:

copy-cursor

It is not possible to display the "copy" cursor in this example if the "default" is set.

PR #529, opened 2021-01-12
Add pivot sphere

Kurtil:
Add a pivot sphere that resizes according to the distance between the camera and the pivot position.
It is possible to enabled or disabled it (default) and change its size and material.

pivotSphere


Amoki:
Don't merge it now.

The sphere geometry breaks with xkt v6 models.

Is there a way to create the sphere in performance mode to allow big coordinates?

Amoki:
I read the section plane plugin's code and I didn't see anything special.
Do you have an idea about why the sphere created like that doesn't render properly with xktv6?
@xeolabs

xeolabs:
@Amoki to position the sphere using double-precision World-space coordinates, you need to split the coordinates into an 32-bit RTC center and a 32-bit offset - see how the SectionPlanePlugin edition control does this the link below.

The idea is that, given a dynamically-changing double-precision 3D world position, we can use [worldtoRTCPos](https://github.com/xeokit/xeokit-sdk/blob/3830d1e3c4d8be2789d1ec3be640b65566bf70df/src/viewer/scene/math/rtcCoords.js#L40) to split that into an RTC center and an offset from that center. The ````Mesh#rtcCenter```` is dynamically set to the center, and ````Mesh#position```` is dynamically set to the offset.

https://github.com/xeokit/xeokit-sdk/blob/master/src/plugins/SectionPlanesPlugin/Control.js#L96

Amoki:
Should be good now!

xeolabs:
I might be misreading, but will this be destroying/creating the sphere mesh as the pivot pos updates?

https://github.com/xeokit/xeokit-sdk/blob/24cf44535f2bd59587a69a594ab6b79bf64f7bbd/src/viewer/scene/CameraControl/lib/controllers/PivotController.js#L128

I think you could just dynamically update ````Mesh#rtcCenter```` and ````Mesh#position```` here, reusing the same ````Mesh```` instance.

Amoki:
The size of the sphere depends on the distance of the pivot point so we can't keep the geometry.
I just fixed a bug where the geometry was contently recreated even if the pivot point was not.

Amoki:
We don't need a very precise sphere, but we can't reduce the resolution:
https://github.com/xeokit/xeokit-sdk/blob/b01f6ddef6e33171dc247a7cced3f001570d2c75/src/viewer/scene/geometry/builders/buildSphereGeometry.js#L72

Is there a reason for these limits?

PR #648, opened 2021-05-10
Distance measurement axes scale bugfix

robin-pham:
Scale: 1
Screen Shot 2021-05-10 at 4 25 23 PM

Scale: 10
Screen Shot 2021-05-10 at 4 25 29 PM

Notice how only the length of the line/wire scales properly, but the axes do not:
```
(if the image is hard to read)
Initial scale
X: 4.69m
Y: 2.83m
Line: 5.48m

10x scale
X: 4.3m
Y: 56.81m
Line: 54.79m
```

 This fixes the calculation so that the axes do the exact same calculation that the line/wire does.


xeolabs:
Thanks!

PR #664, opened 2021-05-26
fix gltf primitives parser

Amoki:
The xkt v6 has the same problem, but I can't submit a PR as the code has been deleted.
We need to maintain backward-compatibility with xkt v6. Do you see a solution or do we must use a fork and don't share the fix with the xeokit community?


xeolabs:
Hmm yes xeokit conversion tools no longer support v6. I suppose you could share a fork with v6 support if that works for you?

Issue #672, opened 2021-06-11
Convert to TypeScript


xeolabs:
Integrate ````tsc````, to automatically build TypeScript source files from the JavaScript source files.

Started in feature branch: https://github.com/xeokit/xeokit-sdk/tree/typescript

xeolabs:
One issue we need to solve: how to configure tsc to work with Pako.js and ZipJS dependencies. See console output below.

````bash
lindsay@zaphod:~/xeokit-sdk$ tsc
src/plugins/NavCubePlugin/CubeTextureCanvas.js:7:1 - error TS9005: Declaration emit for this file requires using private name 'destroy'. An explicit type annotation may unblock declaration emit.

7 function CubeTextureCanvas(viewer, cfg = {}) {
  ~~~~~~~~

src/plugins/XKTLoaderPlugin/parsers/lib/pako.js:1:30 - error TS9005: Declaration emit for this file requires using private name 'Deflate'. An explicit type annotation may unblock declaration emit.

1 /* pako 1.0.10 nodeca/pako */(function(f){if(typeof exports==="object"&&typeof module!=="undefined"){module.exports=f()}else if(typeof define==="function"&&define.amd){define([],f)}else{var g;if(typeof window!=="undefined"){g=window}else if(typeof global!=="undefined"){g=global}else if(typeof self!=="undefined"){g=self}else{g=this}g.pako = f()}})(function(){var define,module,exports;return (function(){function r(e,n,t){function o(i,f){if(!n[i]){if(!e[i]){var c="function"==typeof require&&require;if(!f&&c)return c(i,!0);if(u)return u(i,!0);var a=new Error("Cannot find module '"+i+"'");throw a.code="MODULE_NOT_FOUND",a}var p=n[i]={exports:{}};e[i][0].call(p.exports,function(r){var n=e[i][1][r];return o(n||r)},p,p.exports,r,e,n,t)}return n[i].exports}for(var u="function"==typeof require&&require,i=0;i<t.length;i++)o(t[i]);return o}return r})()({1:[function(require,module,exports){
                               ~

src/plugins/XKTLoaderPlugin/parsers/lib/pako.js:1:30 - error TS9005: Declaration emit for this file requires using private name 'Inflate'. An explicit type annotation may unblock declaration emit.

[...]

PR #713, opened 2021-08-21
fix: fix issue ’Convert to TypeScript #672‘

caiyan1:
Add annotation of /** @private* @return {*}**/  or  /** @private*/ to error files can fix this issue.


caiyan1:
in file of “src/plugins/AxisGizmoPlugin/AxisGizmoPlugin.js”，has other error,I had fixed.

xeolabs:
Thanks!

caiyan1:
Hi，our team need one version which support ts，can you publish one ? Thank you very much !

PR #740, opened 2021-10-02
Move from requestAnimationFrame to setInterval

qs-wang:
Looks we put a lot into requestAnimationFrame which seems not necessary be the optimized option. With my profile switch to setInterval gave better FPS. See attached profiles results.

Note: this changes works quite well on high end laptop, e.g. Apple M1 Pro, which the CPU has enough power. However on a low end PC, the main stream is still be blocked a lot. Other optimization options will be tried.


qs-wang:
Links to the profile files from Firefox.

https://1drv.ms/u/s!Aiq7P3_MfF6li7oOlQheZCMnQFZ_QA?e=uWdFqm

PR #794, opened 2022-01-05
Handle container offsets in distance & angle measurement

Kurtil:
If a container is given to the corresponding plugins, (other than document.body by default), and if the container is not at the top left of the body, the positions may be incorrect.

46801046d276a44f56f547c4ed3b0257


xeolabs:
Now merged: @xeokit/xeokit-sdk 2.2.5-beta-4

PR #800, opened 2022-01-18
fixed mergeNestedFields for las options

DB-EC-Hamburg-BIM:
fixed mergeNestedFields for las options and set DEFAULT_LAS_OPTIONS colordepth to "auto"


DB-EC-Hamburg-BIM:
The PR aims at dist. I was confused. I close it and make a new one in the converter src.
I'm sorry

PR #809, opened 2022-02-17
Fix JS imports

tmarti:
This PR adds the proper `.js` extension to 3 `import` statements.

W/out this change, `xeokit-sdk` cannot be served during its development for example from a `nodejs-express` setup.


xeolabs:
Published in @xeokit/xeokit-sdk@2.2.4

PR #824, opened 2022-03-25
[DataTexturePerformanceModel]: 80% GPU RAM savings!

tmarti:
⚙️ _[WIP] Will keep updating this PR description_ ⚙️

## Introducing `DataTexturePerformanceModel` 🚀 

This is a variation of `PeformanceModel` where all the GPU data is stored in data textures.

GPU memory savings are bigger than 80% (both with instanced and non-instanced geometries) when using the `DataTexturePerformanceModel`.

> 💡This is achieved by using a combination of _creative geometry management techniques_ 💡 
>
>  _**TODO: document the techniques introduced in this PR**_

**It requires `WebGL2` support in the browser**

- https://caniuse.com/webgl2

## Feature toggled! ⚡ 

A toggle is introduced in `XKTLoaderPlugin` to enable the new techniques that optimize GPU RAM usage.

To enable the GPU optimizations, load your `XKT` files with:

```js
xktLoaderPlugin.load ({
    // Same arguments you would use today
    ...,

    // This toggles on the GPU optimizations
    useDataTextures: true,
}); 
```

This should be able to load any XKT file from version `v1` to `v9` (latest today).

### Dynamic Level Of Detail mechanism

When enabled, this extension will try to match a target frame-rate in the Viewer and, if that is not possible, it will keep progressively hiding the most complex objects until the target frame-rate is reached.

When the frame-rate recovers (e.g. when the camera remains steady), it will keep unhiding the previously hidden objects.

The heuristic to know which objects should be hidden first is based on the number of triangles of each object.

By default, the following triangles limits are used:

> `[ 2000, 600, 150, 80, 20 ]`
> - 1st, objects with more than 2000 triangles will be hidden
> - 2nd, objects with more than 600 triangles
> - ...
> - but objects with 20 triangles or less will remain always visible

This is adjustable on a per-model basis:

```js
xktLoaderPlugin.load ({
    // Same arguments you would use today
    ...,

    // This toggles on the GPU optimizations
    useDataTextures: {
        // For this model, set the target frame-rate to 15 fps for LOD mechanism
        targetLodFps: 15,
    },
}); 
```

### View Frustum Culling mechanism

When enabled, this extension will pre-cull all objects outside of the Camera Frustum (this is, objects outside the Field of View of the Camera).

This extension does not cause any visible difference in the render or its quality, but when the camera is close to or inside a model, it's usually the case that only a % of the objects are in the Camera Field of View, and it's in those cases where this extension is able to boost the frame-rate in some cases up to 2x.

> As a technical note, the implementation stands on top of [an efficient `r*tree` based data structure](https://github.com/Eronana/rbush-3d) to boost the Frustum Containment calculations and minimize the number of calculations.
>
> _See the Wikipedia article on `r*trees`: https://en.wikipedia.org/wiki/R*-tree_

In cases where all objects are visible by the Camera, this mechanism adds no noticeable overhead to the frame-rate.

This extension is toggle-able on a per-model basis:

```js
xktLoaderPlugin.load ({
    // Same arguments you would use today
    ...,

    // This toggles on the GPU optimizations
    useDataTextures: {
        // For this model, enable View Frustum Culling
        enableViewFrustumCulling: true,
    },
}); 
```

### Use them together!

```js
xktLoaderPlugin.load ({
    // Same arguments you would use today
    ...,

    // This toggles on the GPU optimizations
    useDataTextures: {
        // For this model, enable both LOD and VFC mechanisms
        targetLodFps: 15,
        enableViewFrustumCulling: true,
    },
}); 
```

## What is working:

- non-instanced geometry rendering (triangles/edges) and picking
- instanced geometry rendering (triangles/edges) and picking
- full illumination support
- out-of-box-support for per-object XYZ offsets
- section plane support
- arbitrary ray-picking
- non-solid geometry rendering (using a novel way to do it)

## What is missing in the `Renderers`

- SAO rendering
- Not all `shader renderers` are migrated. The following ones are missing:
  - `...DepthRenderer.js`
  - `...NormalsRenderer.js`
  - `...OcclusionRenderer.js`
  - `...ShadowRenderer.js`
  - `...ColorQualityRenderer.js`

## What is not covered by the GPU optimizations (yet, to decide if this will be also optimized)

- Lines-based models
- Point-Set-based models

## Some GPU statistics

> `xeokit` is using today around 67 bytes/triangle on the GPU

1. 55.8 M triangles model not using instancing

```
+200k objects
655.23 MB of GPU-RAM
11.15 GPU-bytes/triangle
```

2. 32 M triangles model using instancing

```
+100k objects
125.79 MB of GPU-RAM
3.69 GPU-bytes/triangle
```

### DIFF
--- a/src/plugins/XKTLoaderPlugin/XKTLoaderPlugin.js
+++ b/src/plugins/XKTLoaderPlugin/XKTLoaderPlugin.js
@@ -742,6 +742,7 @@ class XKTLoaderPlugin extends Plugin {
                 isModel: true,
                 origin: params.origin,
                 targetLodFps: params.useDataTextures.targetLodFps || false,
+                enableViewFrustumCulling: params.useDataTextures.enableViewFrustumCulling || false,
             }));
         } else {
             sceneModel = new VBOSceneModel(this.viewer.scene, utils.apply(params, {
--- a/src/plugins/XKTLoaderPlugin/XKTLoaderPlugin.js
+++ b/src/plugins/XKTLoaderPlugin/XKTLoaderPlugin.js
@@ -1,6 +1,6 @@
 import {utils} from "../../viewer/scene/utils.js"
 import {VBOSceneModel} from "../../viewer/scene/models/VBOSceneModel/VBOSceneModel.js";
-import {DataTexturePeformanceModel} from "../../viewer/scene/PerformanceModel/DataTexturePeformanceModel.js";
+import {DataTexturePeformanceModel} from "../../viewer/scene/models/VBOSceneModel/DataTexturePeformanceModel.js";
 import {Plugin} from "../../viewer/Plugin.js";
 import {XKTDefaultDataSource} from "./XKTDefaultDataSource.js";
 import {IFCObjectDefaults} from "../../viewer/metadata/IFCObjectDefaults.js";
--- a/src/plugins/XKTLoaderPlugin/XKTLoaderPlugin.js
+++ b/src/plugins/XKTLoaderPlugin/XKTLoaderPlugin.js
@@ -1,5 +1,6 @@
 import {utils} from "../../viewer/scene/utils.js"
 import {VBOSceneModel} from "../../viewer/scene/models/VBOSceneModel/VBOSceneModel.js";
+import {DataTexturePeformanceModel} from "../../viewer/scene/PerformanceModel/DataTexturePeformanceModel.js";
 import {Plugin} from "../../viewer/Plugin.js";
 import {XKTDefaultDataSource} from "./XKTDefaultDataSource.js";
 import {IFCObjectDefaults} from "../../viewer/metadata/IFCObjectDefaults.js";
@@ -724,6 +725,7 @@ class XKTLoaderPlugin extends Plugin {
      * all geometry instances into batches (````false````), and not use instancing to render them. Setting this ````false```` can significantly
      * improve Viewer performance for models that have excessive geometry reuse, but may also increases the amount of
      * browser and GPU memory used by the model. See [#769](https://github.com/xeokit/xeokit-sdk/issues/769) for more info.
+     * @param {Boolean} [params.useDataTextures=false] When we set this ````true````, an alternative memory representation of object geometry will be used that relies on data textures. At the expense of some rendering performance overhead, this will reduce the used RAM to around 25% respect to setting the option to ````false````.
      * @returns {Entity} Entity representing the model, which will have {@link Entity#isModel} set ````true```` and will be registered by {@link Entity#id} in {@link Scene#models}.
      */
     load(params = {}) {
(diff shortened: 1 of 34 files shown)


tmarti:
🎬 A couple demo videos on what a `data-texture` based approach allows, performance wise.

The tests are done in this model:
- https://github.com/xeokit/xeokit-sdk/tree/master/assets/models/xkt/v7/Lyon
- +73k objects in the scene

Notice how the frame-rate is kept even though continuously  modifying per-object properties 10 times per second!

## Dynamic object coloring 🌈 

73k objects colored each individually 10 times per second!

- more than 700k colorings per second

https://user-images.githubusercontent.com/2405414/160140308-59a7e7d8-0398-45c2-a557-ccb5a722efc6.mov

## Dynamic object coloring and offset 🌈 + 🌊 

73k objects colored each individually 10 times per second!

- more than 700k colorings per second
- more than 700k offsetings per second

https://user-images.githubusercontent.com/2405414/160140790-41cb4caa-0bef-40f4-8ef1-ac8d47dd3d2b.mov

tmarti:
@Amoki @xeolabs here, there is a branch that introduces the VCF and LOD mechanisms to vanilla `PerformanceModel` (w/out `data-textures`):
- https://github.com/tribiahq/xeokit-sdk/tree/test-xeokit-master-vfc-lod

The mechanims there don't need `webgl2` and are toggled a bit differently:

```js
const model = xktLoader.load({
    // as today's xeokit-master
    ...,
    // enable VFC
    enableViewFrustumCulling: true,
    // enable LOD targeting 15 frames per second
    targetLodFps: 15,
 });
```

xeolabs:
Just found one issue @tmarti  

In this test, I'm loading three copies of the Holter Tower, then double-clicking on a roof to fly to it, then when I back way again, the culling does not seem to restore all objects - see example & video:

https://xeokit.github.io/xeokit-sdk/examples/#BIMOffline_XKT_IFC2glTFConverter_RevitSamples_HolterTower_useDataTextures

[Screencast from 21.04.2023 00:08:11.webm](https://user-images.githubusercontent.com/83100/233601693-16663c1e-7c1a-4506-bc68-49aa39a29002.webm)

tmarti:
Super thanks for reporting @xeolabs!

Will check it soon: have the impression that it could be caused, if the same file file is loaded trice, because of repeated id's 🙂.

tmarti:
> Just found one issue @tmarti
> 
> In this test, I'm loading three copies of the Holter Tower, then double-clicking on a roof to fly to it, then when I back way again, the culling does not seem to restore all objects - see example & video:
> 

@xeolabs, this PR solves it: https://github.com/xeokit/xeokit-sdk/pull/1026

tmarti:
@xeolabs please consider also merging this one 🙂 

- https://github.com/xeokit/xeokit-sdk/pull/1027

xeolabs:
@tmarti Performance is very slow on Android 11, Galaxy Tab - testing with [this example](https://xeokit.github.io/xeokit-bim-viewer/app/index.html?projectId=OTCConferenceCenterDataTex&tab=storeys)

tmarti:
> @tmarti Performance is very slow on Android 11, Galaxy Tab - testing with [this example](https://xeokit.github.io/xeokit-bim-viewer/app/index.html?projectId=OTCConferenceCenterDataTex&tab=storeys)

@xeolabs What is the concrete model of the Galaxy Tab and what was its release date?

Ref link: https://en.m.wikipedia.org/wiki/Samsung_Galaxy_Tab_series

xeolabs:
It's a Moto G10 Dual SIM Smartphone (XT2127-2), sorry not a Tab

tmarti:
Then, if you agree, will not implement this as the GPU for the G10 (Adreno 460 within Spandragon 460 SoC) is considered a low-end one by today's standards. Ref: https://nanoreview.net/en/soc/qualcomm-snapdragon-460

xeolabs:
No problem @tmarti - I think it makes sense to expect users to have decent hardware, for our sort of use cases.

tmarti:
Please add this to the list @xeolabs 🎉.

https://github.com/xeokit/xeokit-sdk/pull/1044

This seems to solve the last known problem for data-textures! 🏄

xeolabs:
Nice one @tmarti , merged, thanks!

PR #882, opened 2022-06-29
Fix JS imports by adding the proper `.js` suffix to the `import`-ed files

tmarti:
This PR simply adds the needed suffix to the `import` statements added by latest changes to `xeokit-sdk:master` branch.

This is needed for example so the JS files are found when xeokit-sdk is served from a server using `node/express`.


xeolabs:
Thanks @tmarti , now published in @xeokit/xeokit-sdk 2.3.0-beta-20

PR #923, opened 2022-08-24
Fixed that aspect ratio in imagePlanes only worked for images where width >= height

MaxGeldner:
At the moment the aspect ratio calculation in the ImagePlanes _updatePlaneSizeFromImage function only works for images where width >= height.

This is because we don't know which side is given in the config object when creating an image plane. Depeneding on the side given, the aspect ratio must be calculated with switched values. Because always the bigger side should be given in cfg.size, the aspect ratio used to calculate the other side must be <= 1. With the current implementation this would only work, when the width >= height.

You can see the effects of this in the added [ImagePlane_groundPlane_wide.html](https://github.com/xeokit/xeokit-sdk/compare/master...MaxGeldner:xeokit-sdk:imageplane-aspect-fix?expand=1#diff-c7ea4dd96bbb5b4bc04a67b0a14d5f20287c77f1e938bfa9894b991ea1080f46) example (revert back to the old code to see the wrong version, of course :) ).

This PR changes the calculation of the aspect ratio depending on which side is the bigger one.


xeolabs:
Thanks @MaxGeldner !

PR #960, opened 2022-11-17
Transform to Monorepa with Lerna

kvandake:


xeolabs:
Looking good! We just need to plan a couple of things on our end before merging, since we'll be introducing a bunch of new package names into the SDK now ;)

kvandake:
@xeolabs  We are transforming all packages. Please see it.

xeolabs:
@kvandake looks good to me - I just created a new branch called 'monorepo' - could you please remake this PR against that branch?

In that branch, I've also made all the examples load the binaries from a CDN, instead of the ./dist directory. This will also help us not to disrupt those who are relying on the single binary.

kvandake:
I changed target branch. What are the next steps?

xeolabs:
Next steps on my end:

-merge
- test
- update all the documentation
- publish npm modules, update various links
- update dependent applications

So once I've merged, I'll publish npm packages from that branch as perhaps "beta" and when they're all good, and docs updated, I'll merge into master.

kvandake:
Hello @xeolabs!
When are you planning merge this MR?

xeolabs:
Hi @kvandake, it's coming! I just need to do the next release as a normal repo this week or next, because some users expect it. 

Then after that, I'll apply this PR. I've actually taken your lead on this and used Lerna for xeokit V3 (https://github.com/xeokit/sdk) and it's working well there.

So we're considering Lerna essential and merging is next priority, I expect within the next two weeks.

PR #987, opened 2023-01-13
Added github actions workflows

mlankamp:
Added 2 workflows;
- ci.yaml: runs on pull request and pushes into master
- release.yaml: runs when a new release is created in github (uses the version from the tag created, in form v2.3.8)

Removed dist folder, this is create in the release


mlankamp:
The final push to npm registry needs a secret (NPM_TOKEN) defined in the github settings

xeolabs:
Looks good, thanks!

BTW the examples were loading the binaries from ./dist, but this is good, because I made them fetch from the CDN, which is cleaner I think.

xeolabs:
I'll test this workflow on my end in the next day or so

PR #1017, opened 2023-04-01
Annotation plugin custom z-index for markers & labels

christophe77:
Added custom z-index for Markers & Labels of Annotation plugin.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,18 +1,17 @@
 {
   "name": "@xeokit/xeokit-sdk",
-  "version": "2.3.2",
+  "version": "2.4.0-alpha",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@xeokit/xeokit-sdk",
-      "version": "2.3.2",
+      "version": "2.4.0-alpha",
       "license": "See LICENSE.txt",
       "dependencies": {
         "@loaders.gl/core": "^3.2.6",
         "@loaders.gl/gltf": "^3.2.6",
         "@loaders.gl/las": "^3.2.6",
-        "json-bufferify": "^0.2.2",
         "web-ifc": "^0.0.35"
       },
       "devDependencies": {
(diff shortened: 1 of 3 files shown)


xeolabs:
I needed to roll this one back because it introduced a syntax error and would not compile.
Screenshot from 2023-05-04 14-14-05

Uncaught TypeError: Cannot read properties of undefined (reading '2')
at new AnnotationsPlugin (@xeokit_xeokit-sdk.js?v=73278906:7005:58)
at createAnnotationPlugin (annotations.ts:6:10)
at ifc-player.tsx:137:32
at commitAttachRef (chunk-AXG4LW7R.js?v=09bc691a:15314:15)
at commitLayoutEffects (chunk-AXG4LW7R.js?v=09bc691a:16915:17)
at HTMLUnknownElement.callCallback2 (chunk-AXG4LW7R.js?v=09bc691a:3801:22)
at Object.invokeGuardedCallbackDev (chunk-AXG4LW7R.js?v=09bc691a:3826:24)
at invokeGuardedCallback (chunk-AXG4LW7R.js?v=09bc691a:3860:39)
at commitRootImpl (chunk-AXG4LW7R.js?v=09bc691a:16733:17)
at unstable_runWithPriority (chunk-AXG4LW7R.js?v=09bc691a:348:20)

in this line:
const randomZIndex = 90005 + Math.floor(this._viewPos[2]) + 1;
this._viewPos is not initialized

assaad97:
@xeolabs regarding z-index on annotation marker and label, i noticed that they still exist on the function Annotation::_updatePosition, where a z-index greater than 90005 is enforced.

is this forced z-index is related to a specific use case ?

image

xeolabs:
> @xeolabs regarding z-index on annotation marker and label, i noticed that they still exist on the function Annotation::_updatePosition, where a z-index greater than 90005 is enforced.
> 
> is this forced z-index is related to a specific use case ?

No not really - it's just a hack to ensure that annotations always float about things like HTML elements used to represent measurements, a guess at a large enough z-index value.

PR #1026, opened 2023-04-24
data-textures: make `VFC` mechanism support non-identity `model.worldMatrix`

tmarti:
This fixes a problem where View Frustum Culling would not work well for models loaded with e.g. the `cfg.position` parameter different from the zero vector.

This fix is based on the fact that it is the same (during `VFC`):

- to pre-process the AABB's so they include the `model.worldMatrix` and this makes the culling consider final world-AABB's. Then the clipping happens in world-space.
- to instead transform the camera `eye`/`look` with the inverse transform, so the AABB's are kept the same. Then the clipping happens in model-space.


tmarti:
Solves https://github.com/xeokit/xeokit-sdk/pull/824#issuecomment-1517554791

PR #1027, opened 2023-04-25
Additional data-textures feature toggles

tmarti:
## Description

A couple loading options have been added to `XKTLoaderPlugin` when using data-textures:

```js
const model = xktLoaderPlugin.load({
    ...,
    useDataTextures: {
        ...,

        // optional: will enable/disable welding vertex positions when loading geometry,
        disableVertexWelding: true/false, // default: false

        // optional: will enable/disable adaptive demotion of triangles' indices bitness then loading geometry
        disableIndexRebucketing: true/false, // default: false
    }
});
```

- `useDataTextures.disableVertexWelding` is interesting in case the XKT files already contain welded vertices positions. In that case setting this value to `true` will speed up the loading of the XKT files.
- `useDataTextures.disableIndexRebucketing` is interesting to set it to `true` in case we want to save further load time, at the expense of slightly increased VRAM usage.

## Backwards compatible

The default values of the new options are made in such a way that the current behaviour in `main` branch is kept.

The default values correspond to maximum VRAM savings.

## Independent of other `data-texture` feature toggles

The two options introduced in this PR can be toggled separately, and they're also toggleable independent of the values of the already existing feature toggles for `data-textures`:

- `useDataTextures.targetLodFps`
- `useDataTextures.enableViewFrustumCulling`

### DIFF
--- a/src/plugins/XKTLoaderPlugin/XKTLoaderPlugin.js
+++ b/src/plugins/XKTLoaderPlugin/XKTLoaderPlugin.js
@@ -744,6 +744,7 @@ class XKTLoaderPlugin extends Plugin {
                 targetLodFps: params.useDataTextures.targetLodFps || false,
                 enableViewFrustumCulling: params.useDataTextures.enableViewFrustumCulling || false,
                 disableVertexWelding: params.useDataTextures.disableVertexWelding || false,
+                disableIndexRebucketing: params.useDataTextures.disableIndexRebucketing || false,
             }));
         } else {
             sceneModel = new VBOSceneModel(this.viewer.scene, utils.apply(params, {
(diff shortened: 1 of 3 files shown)


tmarti:
Also, in case your're NOT welding positions, it's highly recommendable to enable rebucketting.

In my tests for mid-size models (starting at 20M triangles):

- turning off rebucketting and without welding increases VRAM usage by around 60% extra
- turning off rebucketting but with welding icreases VRAM usage instead by around 10% extra

But it also depends on the case: I'm using XKT files with a tremendeous amount of instancing:
- I use to handle XKT files where >90% of the geometry is instanced
- and the memory consumption w/out rebucketting gets worse with lower levels of instancing.

For all this it depends a bit on the concrete case, so I just encourage you to play with the two toggles in this PR and find the balance between load time and memory usage that works best for you.

The default values are set for minimum memory consumption, at the expense of around 30% penalty in load times.

PR #1075, opened 2023-06-10
data-tex: reduce number of per-portion matrices

tmarti:
Here, instead of keeping in the VRAM, per object...

- one texture for the positions-decode matrix
- one texture for the entity matrix

... we just store the result of concatenating both.

In models using a high level of instancing, this saves a good amount of VRAM due to textures.

For example, in an internal test loading a grid with 16x16 Holter Towers...
- 435 geometries
- but 1M portions
- (2400x geometry re-usage factor)

... the total texture VRAM usage goes down from to **206.73 MB** to **139.95 MB**


xeolabs (approved):
Foo

PR #1085, opened 2023-06-22
Add stats typings

LukasKairnial:
Hi, I noticed the [stats](https://github.com/xeokit/xeokit-sdk/blob/a2bf178671392f0703b2f56918d8ccf9c914ed2c/src/viewer/scene/stats.js) object was missing typings, which prevent correct Typescript import.


xeolabs:
Thanks @LukasKairnial !

PR #1107, opened 2023-07-19
Prevent viewer getSnapshot to add canvas on document

Kurtil:
[This change](https://github.com/xeokit/xeokit-sdk/commit/78ff80c4db6ec012b9e3753c7241e7e5ae0f94f5#diff-3986a1ed352df2aeac30a1e1db927c1a2285e4ce219ec61dfa57406f01ef4983R349) add a breaking change in the `viewer.getSnapshot()` API. I use the `viewer` with a distance measurement plugin with a specified container that is not `document.body`. But since this change, when a get a snapshot of the viewer, a canvas is also added on the `document.body`. I propose to add a `capture` parameter on `getShapshot` to prevent this behaviour if not requested (must be set to `true` if captured `canvas` are desired).


Kurtil:
This may not be what we need :P @Amoki will make another PR related to the issue described here.

PR #1109, opened 2023-07-20
[WIP] Upgrade VBO Scene Model rendering performances

Kurtil:


Kurtil:
Arf... again my bad it was for the fork :/

xeolabs:
I am intrigued though! Do you have some rendering performance tweaks incoming? If so, could you give me a quick summary?

Kurtil:
> @Kurtil I am intrigued though! Do you have some rendering performance tweaks incoming? If so, could you give me a quick summary?

Yes! Recording a flame graph on a huge model, we saw that calling `matrix4fv` too much seems to reduce performances on picking. As all `triangle` renderers have relatively the same matrices in their shaders, creating a Uniform Block Buffer regrouping all main matrices (world, view, proj, decodePosition, worldNormal, viewNormal) should reduce WebGL calls and increase performance.
To go further, I'll try to reduce WebGL calls even more by using `Vertex Array Object` to prevent calling [`ArrayBuffer.bindArrayBuffer`](https://github.com/xeokit/xeokit-sdk/blob/master/src/viewer/scene/webgl/Attribute.js#L13) (is calls internally `enableVertexAttribArray`, `vertexAttribPointer` & `bindBuffer`) on each attributes for each draw.
If the performance boost is convincing, I will let you know with a PR :)

Kurtil:
See [here](https://github.com/xeokit/xeokit-sdk/pull/1113) for the full PR.

Issue #1076, opened 2023-06-11
Make picking (and camera navigation) faster by using a 1x1 viewport


xeolabs:
Currently, our GPU-assisted picking renders the object pick colors and depths to a full-sized set of buffers, equal to the size of the canvas.

This is wasteful, and I think we would get a noticeable improvement in responsiveness if we instead render pick colors and depths to a 1x1 sized buffer, which would also involve using a custom projection, align the camera along the ray passing through the front plane at the clicked canvas coordinates etc.

 The 1x1 viewpoint technique is currently used in vertex snapping: https://github.com/xeokit/xeokit-sdk/blob/master/src/viewer/scene/webgl/Renderer.js#L1358

tmarti:
Check how the clip-space remapping is done in this PR:
- https://github.com/xeokit/xeokit-sdk/pull/1062

Pro-tip: try to avoid using a custom view/projection matrix 🙂, there is a distortion of perspective towards the sides of the camera frustum if you instead try to temporarily set the camera eye/look/up to get a ray algined with the canvas coordinate.

In the mentioned PR, I ended up implementing the clip-space remapping technique (so the shader does not suffer from that perspective distortion).

I'm attenting personal things those days, but in case you're interested I'm looking forward to write a small article about the technique in the PR in some days/weeks 🙂

tmarti:
Further explanation of clip-space remapping: https://github.com/tmarti/graphics-research/blob/main/articles/webgl2-gpu-based-snapping.md#clip-space-remapping

Kurtil:
1x1 picking viewport using clip-space remapping PR is merged. This issue can be closed.

PR #1113, opened 2023-08-01
VBO Scene Model Renderers refactoring and performance improvement

Kurtil:
VBOSceneModel renderers contain a lot of duplicated lines. This PR aims to reduce duplicated lines by regrouping the logic into a new abstract class [`VBOSceneModelRenderer`](https://github.com/xeokit/xeokit-sdk/pull/1113/files#diff-4300f821d4ee6f15d73b037e60c7173f6c24bdb0390339c0227bfca2f6c51eed).

`_allocate`, `drawLayer` & `_bindProgram` are now the same for each renderers. The only differences are some parameters or the availability of the uniforms/attributes. As [`Program`](https://github.com/xeokit/xeokit-sdk/blob/master/src/viewer/scene/webgl/Program.js) attributes & uniforms depends of the corresponding shaders, it is simple to know if the current renderer needs to provide a data for the attribute or uniform.

Renderers with this PR are a VBOSceneModelRenderer ( with minimal config like one or two booleans)  + a pair of shaders.

In addition of the refactoring, this PR aims to increase performances using two WebGL APIs:

- Uniform Block Buffer for basic matrices (world, view, proj, positionDecode, worldNormal, viewNormal)
- Vertex Array Object by layer

Indeed, the first goal of these changes was to increase performance for picking. I had laggy performance with a large model while picking to highlight. With the chrome performance snapshot tab, I saw that `uniformMatrix4v` seems to be the main bottleneck. To prevent this, I used UniformBlockBuffer to provide matrices to the programs. After this change, the bottleneck was a lot of `attributeDivisor` & `vertexAttributePointer` calls. To limit these calls, I add [VAOs](https://developer.mozilla.org/en-US/docs/Web/API/WebGLVertexArrayObject) in [this commit](https://github.com/xeokit/xeokit-sdk/pull/1113/commits/ee77b4211c8b101af42fd3bc6179054b84ea3a18).

Now the main bottleneck is readpixel :) something like [this](https://github.com/xeokit/xeokit-sdk/issues/1076) could benefits this PR.

All scene representation VBOSceneModel * are still working ISO however it needed to be benchmarked.

Further improvement: The [set of the UniformBlockBuffer for matrices](https://github.com/xeokit/xeokit-sdk/pull/1113/files#diff-4300f821d4ee6f15d73b037e60c7173f6c24bdb0390339c0227bfca2f6c51eedR454-R470) could be cached if camera did not change for a entire frame rendering.


xeolabs:
@Kurtil excellent, thanks, this refactor was exactly what's needed here. I had consciously duplicated things as a first step to avoid initial complexity, with the intention of pushing duplicate code down into a "template base class" later on, once it worked, just like this, but alas never got around to it! 

Also, in case it's interesting, I ended up pushing common code into a base class in the [xeokit V3 rewrite](https://github.com/xeokit/sdk/tree/main/packages/webglrenderer) (but that code doesn't work yet, maybe later this year).

PR #1121, opened 2023-08-14
Use a `ResizeObserver` to detect changes canvas size

tmarti:
## Description

Today, in each scene `tick`, the canvas and window sizes are compared to their previous values to detect if the pixel-size of the canvas needs to be updated. By doing that, calls happen to `.clientWidth` and `.clientHeight` properties, and this can trigger DOM layout calculations per-`tick`, which adds overhead to the rendering process (according to Chrome Profiler).

Instead, this PR pretty much simplifies the mechanism and delegates the knowledge about if the canvas resized to a `ResizeObserver`, which avoid to kick any extra DOM layout calculation: the code gets simpler and more performant.

## Bonus track

Also, when the `Perspective` matrix is updated, the class will make sure to mark its `Camera` as "dirty" so it updates the matrix. This is needed in some cases so the camera gets aware that the x/y axis-ratio has changed, as the aspect ratio is handled on the projection matrix.

## Reference

`ResizeObserver` is supported on all major browsers since around 3 years ago: 

- https://caniuse.com/resizeobserver

### DIFF
--- a/src/viewer/scene/camera/Perspective.js
+++ b/src/viewer/scene/camera/Perspective.js
@@ -78,6 +78,8 @@ class Perspective extends Component {
 
         this.glRedraw();
 
+        this.camera._updateScheduled = true;
+
         this.fire("matrix", this._state.matrix);
     }
 
--- a/src/viewer/scene/canvas/Canvas.js
+++ b/src/viewer/scene/canvas/Canvas.js
@@ -72,9 +72,6 @@ class Canvas extends Component {
          */
         this.transparent = !!cfg.transparent;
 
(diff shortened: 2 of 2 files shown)


xeolabs:
@tmarti I wonder if this BIMViewer issue is somehow related: https://github.com/xeokit/xeokit-bim-viewer/issues/155

tmarti:
I think so; for the perspective case this commit in this PR should solve it:

- https://github.com/xeokit/xeokit-sdk/pull/1121/commits/8f73899f0011f2f0dcbf296ff37212209c9cafdd

PR #1158, opened 2023-10-04
feature: add dual snap mode => vertex + edge

Kurtil:
This PR adds the ability to snap vertex AND edge. If both are found, vertex are snapped with higher priority. The main purpose of this changes is to do this dual snap picking with only one depth buffer initialisation to improve performance in comparaison of doing two snap picking simultaneously with different snap mode.

dualSnap

**API changes** : (BEAKING CHANGES)

- `snapMode` is only used as `frameCtx` property.
- CameraControl now has `snapVertex: boolean` & `snapEdge: boolean` properties instead of `snapMode`. (In case of both are `false`, the snapPick falls back to a regular pick with `canvasPos` and `pickSurface: true`)
- scene.snapPick accept `snapVertex: boolean` & `snapEdge: boolean` arguments instead of `snapMode`.
- `pickResult` has `snapType`: null|"vertex"|"edge" property instead of `snappedToVertex: boolean` & `snappedToEdge: boolean`.

**Implementation** :

Only `Renderer` logic, no shader update. No changes for mono type snap (only vertex or only edge).

On dual type snap, `frameCtx.snapPickLayerParams` stores empty x 1 then the edge layer params first, then empty x 1, then the vertex layer params. The empty x 1 is because the `w` of the drawn snap color is 0 if no snap, and if !== 0, represents the layer params number. The second empty x 1 is to make this line works fine (done by the `frameCtx.snapPickLayerNumber++` between the two `snapPickDrawSnapDepths` calls):

```js
const pickedLayerParmasSurface = layerParamsSurface[Math.abs(pickResultMiddleXY[3]) % layerParamsSurface.length];
```

Indeed, if `layerParamsSurface` has a length of 10, pickResultMiddleXY[3] can be 12 if a vertex snap was drawn but thanks to the % it falls back to the 2 index and layerParamsSurface[foundIndex] is not `undefined`.

Then, this line allow to know if the snap pick is a vertex pick or an edge pick:

```js
const isVertex = snapVertex && snapEdge ? snapPickResultArray[i + 3] > layerParamsSnap.length / 2 : snapVertex,
```

Has edge pick are stored on the first half of the `layerParamsSnap` array and vertex pick on the last half.

**To go further**

In case of a snapPick in center position, the previous implementation got the `pickedLayerParmasSurface` from the `layerParamsSurface` (see [here](https://github.com/xeokit/xeokit-sdk/blob/master/src/viewer/scene/webgl/Renderer.js#L1357)) array which is not correct as snap pick params are stored in the `layerParamsSnap` array. But if this is fine, the `layerParamsSnap` may not be really needed as it is the same values as the `layerParamsSurface` array. It may be interesting to simplify this code to only use the `layerParamsSurface` array.

### DIFF
--- a/src/plugins/DistanceMeasurementsPlugin/DistanceMeasurementsControl.js
+++ b/src/plugins/DistanceMeasurementsPlugin/DistanceMeasurementsControl.js
@@ -144,8 +144,11 @@ class DistanceMeasurementsControl extends Component {
                 this.markerDiv.style.marginLeft = `${event.canvasPos[0] - 5}px`;
                 this.markerDiv.style.marginTop = `${event.canvasPos[1] - 5}px`;
                 this.markerDiv.style.background = "pink";
-                if (event.snappedToVertex || event.snappedToEdge) {
+                if (event.snapType === "vertex") {
                     this.markerDiv.style.background = "greenyellow";
+                    this.markerDiv.style.border ="2px solid green";
+                } else if (event.snapType === "edge") {
+                    this.markerDiv.style.background = "lightblue";
                     this.markerDiv.style.border ="2px solid green";
                 } else {
                     this.markerDiv.style.background = "pink";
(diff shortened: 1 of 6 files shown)


xeolabs:
Makes perfect sense, thanks @Kurtil

PR #1159, opened 2023-10-04
Update node.js import to make them es compatible

Kurtil:
While testing new stuffs on the repo, I use to test examples with a web server (vscode live server) while updating the import of the xeokit sdk from `dist` to `src`. However, a lot of import are not web compatible but node.js compatible. Because of that, it is a pain to serve the examples without a bundler.

This PR aims to ensure the maximum of import are web compatible (some of theme like html2canvas needs to be processed by a bundler or commented to let the viewer start on the browser with the dev server).

### DIFF
--- a/src/plugins/GLTFLoaderPlugin/GLTFLoaderPlugin.js
+++ b/src/plugins/GLTFLoaderPlugin/GLTFLoaderPlugin.js
@@ -2,7 +2,7 @@ import {Plugin, SceneModel, utils} from "../../viewer/index.js"
 import {GLTFSceneModelLoader} from "./GLTFSceneModelLoader.js";
 
 import {GLTFDefaultDataSource} from "./GLTFDefaultDataSource.js";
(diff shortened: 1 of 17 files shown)


Kurtil:
🎉 @xeolabs no need to comment the html2canvas import of the Viewer.js file if the browser supports [importmap](https://developer.mozilla.org/fr/docs/Web/HTML/Element/script/type/importmap). I had these lines on the html head of the desired examples and it works perfectly:

```html
    <script type="importmap">
        {
          "imports": {
            "html2canvas/dist/html2canvas.esm.js": "https://unpkg.com/html2canvas@1.4.1/dist/html2canvas.esm.js"
          }
        }
    </script>
```

Thanks to this PR and the lines above, (+ update the import of the xeokit-sdk from dist to src), it is possible to dev on xeokit with a minimal dev environment (a dev server) ! :D

Issue #1105, opened 2023-07-18
Ray picking precision


Kurtil:
I'm developing a plugin to measure distances between objects.

![bbbf204df94bada3ce9e8e0efc994eb1](https://github.com/xeokit/xeokit-sdk/assets/22523482/d0614978-6137-4513-9a31-7852496ea00f)

However, ray picking is not precise enough to consider my plugin as usable.

**First**, it is not possible to ray picking precision surface with `matrix` instead of `origin` & `direction`. Indeed, `worldRayOrigin` & `worldRayDir` are not set if a matrix is provided in pick `params`. See [here](https://github.com/xeokit/xeokit-sdk/blob/master/src/viewer/scene/webgl/Renderer.js#L1038). Is this an oversight or it is intended to does not work with matrix ?

**Second**, I used picking with matrix because it seems to be more precise. Indeed, the [matrix ray picking example](https://xeokit.github.io/xeokit-sdk/examples/#picking_ray_matrix) seems way more *stable* than the [surface ray picking example](https://xeokit.github.io/xeokit-sdk/examples/#picking_ray_surface). Comparing the code of both examples and the code on the Renderer.js, I saw that the only difference to go from `origin` & `direction` to `matrix` is [this part](https://github.com/xeokit/xeokit-sdk/blob/master/src/viewer/scene/webgl/Renderer.js#L971-L976):

```js
randomVec3[0] = Math.random();
randomVec3[1] = Math.random();
randomVec3[2] = Math.random();

math.normalizeVec3(randomVec3);
math.cross3Vec3(worldRayDir, randomVec3, up);
```

Maybe these lines are responsible for the lack of precision when picking with `origin` & `direction`...

Measuring the Duplex roof is a good place to show the ray picking limitations:

**No Precision**
- `origin` & `direction`:
![8b0b929e0155c8a0ac031f874d734d71](https://github.com/xeokit/xeokit-sdk/assets/22523482/eabc8f61-8fb2-460d-a6f0-2b5698ddd470)
- `matrix`:
![6776d57f2d08e712a143fd823ea901bf](https://github.com/xeokit/xeokit-sdk/assets/22523482/bf51577e-e888-4c61-9e74-f7c9458a3af7)

I seems better with matrix but in both no precision cases, the measure stops like 10cm before the wall...

For this kind of plugin, precision is key, precision picking needs to be enabled.

**Precision**
- `origin` & `direction`:
[...]

PR #1168, opened 2023-10-09
1 x 1 picking viewport

Kurtil:
Picking with a 1 x 1 viewport. I used the clip position remapping method used by @tmarti for the snap picking.

WARNING : ray picking is broken and needs to be fixed before merging this PR! :(

Before :

rayPickSurface-before

After (with this PR merged) :

rayPickSurface-after

It seems the before is not perfectly accurate. Maybe these changes accentuate the small inaccuracies...


Kurtil:
Ray picking precision issue https://github.com/xeokit/xeokit-sdk/issues/1105.

xeolabs:
@Kurtil is BIMData depending on ray-picking functionality (with origin->ray) ? I don't even know of any users who are actually depending on ray-picking, so if that can't be fixed, I'd be in favour of removing ray-picking altogether, if that stands in the way of faster surface picking with canvas coordinates (enabled by this PR). Fast model interaction is always our top priority (and canvas surface picking is a big part of that, since it's used by camera interaction).

xeolabs:
@Kurtil I've fixed the precision issue while doing a merge conflict fix: https://github.com/xeokit/xeokit-sdk/pull/1168/commits/e9549ca02df787e444831ed1d2f5e2722ac887c9

tmarti:
@xeolabs at Tribia we're also depending on ray picking for laser mesurements, plus some custom functionalities for obtaining the ifc-storey and ifc-space where the camera is! 🙂

Issue #1162, opened 2023-10-05
Z-fighting while snap picking


Kurtil:
While developing the dual snap mode (edge + vertex), I realised that cursor isn't always snapping the way it is supposed to do. To debug snapping, I draw the texture used in the snap pick frame buffer on the left of the screen with red colore for edge & green colore for vertex (snap radius 100):

Mid distance snapping:

![MidSnap](https://github.com/xeokit/xeokit-sdk/assets/22523482/d8b6f9e8-af08-4b7c-81fd-202c4d76097c)

Close distance snapping:

![CloseSnap](https://github.com/xeokit/xeokit-sdk/assets/22523482/a7001fc1-d4f7-457d-937a-416822376f3e)

Far distance snapping:

![FarSnap](https://github.com/xeokit/xeokit-sdk/assets/22523482/7ea7e7c6-9753-4c99-b623-07e9889e0cd9)

Red lines are laggy and sometimes green dot disappear. It seems to be Z-Fighting because this is the content of the texture with a mid distance snapping with depth test disabled :

<img width="1039" alt="Screenshot 2023-10-05 at 17 01 37" src="https://github.com/xeokit/xeokit-sdk/assets/22523482/ebb0d06e-6119-481e-b7ed-c953cfb6800a">

No laggy lines or missing dot.

**Exploring the depth problem :**

Logarithmic depth buffer is [always enabled for snap picking](https://github.com/xeokit/xeokit-sdk/blob/master/src/viewer/scene/model/vbo/snapBatching/SnapBatchingDepthRenderer.js#L11) to improve occlusion accuracy at distance. But what really seems to mess up with occlusion accuracy at distance seems to be [the small z offset added in the shader](https://github.com/xeokit/xeokit-sdk/blob/master/src/viewer/scene/model/vbo/snapBatching/SnapBatchingDepthBufInitRenderer.js#L220):

```js
src.push("clipPos.z += 0.0001;"); // small Z offset
```

I test with different configuration (Logarithmic depth buffer ON/OFF, zOffset ON/OFF) :

<img width="883" alt="Screenshot 2023-10-05 at 16 51 52" src="https://github.com/xeokit/xeokit-sdk/assets/22523482/7dd78628-39e7-457b-9a5a-d44589de8889">

Far distance :

[...]

PR #1195, opened 2023-10-25
Use gradient technique to depth-init snapping buffers

tmarti:
Solves https://github.com/xeokit/xeokit-sdk/issues/1162

### DIFF
--- a/src/viewer/scene/model/dtx/triangles/renderers/TrianglesDataTextureSnapDepthBufInitRenderer.js
+++ b/src/viewer/scene/model/dtx/triangles/renderers/TrianglesDataTextureSnapDepthBufInitRenderer.js
@@ -371,7 +371,6 @@ export class TrianglesDataTextureSnapDepthBufInitRenderer {
         src.push("float tmp = clipPos.w;")
         src.push("clipPos.xyzw /= tmp;")
         src.push("clipPos.xy = remapClipPos(clipPos.xy);");
-        src.push("clipPos.z += 0.0001;"); // small Z offset
         src.push("clipPos.xyzw *= tmp;")
         if (SNAPPING_LOG_DEPTH_BUF_ENABLED) {
             src.push("vFragDepth = 1.0 + clipPos.w;");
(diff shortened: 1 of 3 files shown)


xeolabs:
> Still, some problems with the following example page:
> 
> ```
> examples/scenemodel/snapToVertex_dtx_instancing_triangles.html
> ```
> 
> Will keep digging here

Nice one @tmarti - how's it looking with the other snapping examples? Worth merging at this point, or would you like to refine more?

tmarti:
Can you please answer with a link to the other examples?

tmarti:
But yes, should be safe to merge, and maybe @Kurtil (the OP) you can confirm it solves the problems you were observing 🙂

xeolabs:
> Can you please answer with a link to the other examples?

Scroll down to "Snap Picking" for a complete set of tests for snapping:

https://xeokit.github.io/xeokit-sdk/examples/scenemodel/

xeolabs:
> Still, some problems with the following example page:
> 
> ```
> examples/scenemodel/snapToVertex_dtx_instancing_triangles.html
> ```
> 
> Will keep digging here

Note that the same issue also happens in master, prior to merging this PR, so is not caused by this PR. Therefore, I'll merge.

Could it somehow be related to triangle edges that are only adjacent to a single triangle?

[Screencast from 25.10.2023 22:50:07.webm](https://github.com/xeokit/xeokit-sdk/assets/83100/fa4244dc-0ab5-4882-8ede-558f3d629484)

tmarti:
> Wondering, could it somehow be related to triangle edges that are only adjacent to a single triangle?

It indeed looks like that’s the case. Might need to debug a bit that shader.

tmarti:
@xeolabs, would you see with good eyes introducing @Kurtil's snapping-buffer-debugger code somehow hidden under a toggle?

Of course always disabled by default, but think it would be really nice to keep his precious "visual debugger" in case there are problems again in the future with snapping 🙂.

It could come disabled by default and maybe (not necessarily this way, just an example):

```js
const v = new Viewer({ 
    // ...
    debugOptions: {
        showSnapBufferInspector: true|false, // default false
    }
});
```

Issue #1263, opened 2023-11-24
Dynamic zoom speed with distance from camera doesn't work anymore


Amoki:
**Describe the bug**
Zoom speed is constant and doesn't slow down when the focused object is near.
Can't it come from https://github.com/xeokit/xeokit-sdk/pull/1261?

**To Reproduce**
1. Go in a building
2. Zoom
3. Camera goes too fast

xeolabs:
Probably. I didn't notice that when testing. Unfortunately I'm out of time to fix this until Sunday.

Amoki:
The object picking is also bugged `viewer.cameraControl.on("picked")` is not emited. I tried and both problems come from this PR

tmarti:
That's unfortunate! Will fix tomorrow morning 🙂👌

Amoki:
No worries!
I totally can wait

tmarti:
@Amoki: also, do you have some sample HTML page for the case of `viewer.cameraControl.on("picked")`

xeolabs:
@tmarti I can easily do a rollback on this no worries, no pressure man. 

I just had a little look; basically the issue is that there are some handlers within CameraControl that expect a pick to happen synchronously when PickController.update() is called, where the handlers then check for results on PickController immediately after calling. So, when that is asynchronous, the results won't be there immediately after, and we'll get some missing pick events.

An idea: perhaps `PickController.update()` could cache its pick results until the next tick? So the first time it's called after the last tick, do an immediate pick, cache the results and return. Then if `update()` is called again, and a new tick hasn't happened yet, return the same results. Then as soon as a new tick happens, clear the results etc.

This would assume that the view does not update in any way between multiple calls to update() within the current render frame, which I believe is a safe assumption, since a render only happens once per tick.

The cache would be valid as long as the combination of values of these don't change since the last call to `update():

* PickController.pickCursorPos
* PickController.schedulePickEntity
* PickController.schedulePickSurface
* PickController.scheduleSnapOrPick

We could hash those. On each update() if the same values are on these since the last time, cached results would still be valid.

CameraControl.fireEvents() can remain the way it is, no problem there AFAIK. 

Also a quick hack for now, if you run out of time, could be to make PickController.update() method call PickController.runUpdate().

xeolabs:
@tmarti check it out, maybe this does the trick: https://github.com/xeokit/xeokit-sdk/pull/1264

[...]

PR #1265, opened 2023-11-26
OPTIMIZATION: tickify `mousemove` and `mousewheel` events

tmarti:
## Description

TLDS; this PR solves the bug introduced in #1261 plus it keeps providing a way to "debounce" `mousemove` and `mousewheel` events

## The problem with #1261 

The PR #1261 tried to fix two problems:

- i) ✅ a bug where some kind of "race condition" happened while taking distance measurements. This bug was solved in https://github.com/xeokit/xeokit-sdk/pull/1261/commits/a92b88a75f83705860a9996c8f79ef0056d6d0ac. **This fix is still valid**. ✅ 

- ii) ❌ a problem where a massive amount of `mousemove` events handled by the event handler in `MousePickHandler` was causing a huge lag for example when creating distance measurements (https://github.com/xeokit/xeokit-sdk/pull/1261/commits/0fc7266fc6aaa6ecc53a365af19c93dfa170e033). This is what introduced the bug discovered by @Amoki in https://github.com/xeokit/xeokit-sdk/issues/1263 ❌ 

## The reason for the bug introduced in #1261 

That PR assumed that the state of the `PickController` was consumed in an async way. What the PR did was to delay the execution of the `PickController.update()` method until next tick frame.

But, turns out certain functionalities like the input-zoom-handler was doing a sync consumption of the `PickController`. So, as the PR refactored the `PickController` in such a way that the pick result was available only after next-scene tick, clients of `PickController`that expected an immediate picking result stopped working. Those are the two bugs found by @Amoki, and could be more out there hidden caused by that PR! 😱 

## What performance problem did #1261 try to fix

To show what was that problem, here I include an instrumented version of `Renderer.js`.

It will console-log the number of GPU-picks per scene-tick:

[Renderer.js.patch (in diff format)](https://github.com/xeokit/xeokit-sdk/files/13466321/Renderer.js.patch)

This video shows the console output when this patched renderer runs while creating distance measurments. And the model is the really simple `Duplex` model!

https://github.com/xeokit/xeokit-sdk/assets/2405414/03dd7469-08f2-4d86-8aa3-fda3566b4046

As you can see, in some scene ticks (this is, in a single frame update) more than 200 GPU-pickings were triggered!

The result of this was really laggy measurement creation. And, #1261 tried to fix it in the wrong way.

## How this PR approaches the same problem

The problem with the previous PR, is that it assumed that the consumers of `PickController` were consuming it in an async way. But, this assumption is not true either for async or sync consumption. It should be respected just any possible way it's consumed to avoid specific code refactors.

Instead, this PR starting point is the following: **what we need to run maximum once per-tick, is now any internal piece of code but the handlers themselves**!

This is, the problem in the first place is that during a single scene tick a big number of mouse-move events are consumed by the same handler.

So...

## Introducing `tickify`-cation

This PR introduces a new `Scene.tickify(function)` method!

The core idea, is that any function can be passed to this method will be wrapped so it's run a maximum of ONE time per scene tick.

```js
// Tickify the function
const tickified = scene.tickify(
    /* some function */
);

// Now, when we invoke tickified...
tickified();

// ... we are sure that the code in the original function
// will be invoked maximum once per scene tick!

// e.g. no matter the loop, this will be invoked just once in next scene tick
for (let i = 0; i < 1000; i++) {
    tickified();
}
```

## Intented usage

Today, some event handlers do the following:

```js
canvas.addEventListener("mousemove", /* some function */);
```

Using `tickify(...)`, this would be converted into:

```js
const tickified = scene.tickify(/* some function */);

canvas.addEventListener("mousemove", tickified);
```

## How this differs from PR #1261 

This approach does not change in any way how the `PickController`, or any other method interally used by event handlers, work. So, does not touch on any async/sync assumption. This also removes the bug surface of #1261.

Instead, it just adds a trigger-condition to the handlers. And the trigger-condition just makes sure they're invoked once per time at maximum. And, it allows the event handlers to use same internal code and in same way.

Or course, it introduces a delay of one scene tick for when the handlers will be executed. But that should not matter as mouse event triggering is asynchronous in nature until it reaches the Js code!

## The result

Running same patched `Renderer.js` on this PR's branch, the result is now a much lower number of picks per scene-tick:

Notice how now the number of picks-per-tick is constant and way lower than in the original scenario 🚀 

https://github.com/xeokit/xeokit-sdk/assets/2405414/5491968c-5c09-4c68-905a-161dee13c403

## What is missing in this PR

The `tickify`method in `Scene` class keeps track of the tickified functions, so it's possible to `untickify` the functions later. And this should be done also when unsubscribing the DOM event listeners.

This clean-up part is not implemented yet by the consumers of `Scene.tickify()` that have been refactored in this PR.

### DIFF
--- a/src/viewer/scene/scene/Scene.js
+++ b/src/viewer/scene/scene/Scene.js
@@ -352,6 +352,11 @@ class Scene extends Component {
             throw "Mandatory config expected: valid canvasId or canvasElement";
         }
 
+        /**
+         * @type {{[key: string]: {wrapperFunc: Function, tickSubId: string}}}
+         */
+        this._tickifiedFunctions = {};
+
         const transparent = (!!cfg.transparent);
         const alphaDepthMask = (!!cfg.alphaDepthMask);
 
@@ -2560,6 +2565,63 @@ class Scene extends Component {
             }
         }
         return changed;
+    }
+
+    /**
+     * This method will "tickify" the provided `cb` function.
+     * 
+     * This means, the function will be wrapped so:
+     * 
+     * - it runs time-aligned to scene ticks
+     * - it runs maximum once per scene-tick
+     * 
+     * @param {Function} cb The function to tickify
+     * @returns {Function)}
+     */
+    tickify(cb) {
+        const cbString = cb.toString();
+
+        /**
+         * Check if the function is already tickified, and if so return the cached one.
+         */
+        if (cbString in this._tickifiedFunctions) {
+            return this._tickifiedFunctions[cbString].wrapperFunc;
+        }
+
+        let alreadyRun = 0;
+        let needToRun = 0;
+
+        let lastArgs;
+
+        /**
+         * The provided `cb` function is replaced with a "set-dirty" function
+         * 
+         * @type {Function}
+         */
+        const wrapperFunc = function (...args) {
+            lastArgs = args;
+            needToRun++;
+        };
+
+        /**
+         * An each scene tick, if the "dirty-flag" is set, run the `cb` function.
+         * 
+         * This will make it run time-aligned to the scene tick.
+         */
+        const tickSubId = this.on("tick", () => {
+            const tmp = needToRun;
+            if (tmp > alreadyRun) {
+                alreadyRun = tmp;
+                cb(...lastArgs);
+            }
+        });
+
+        /**
+         * And, store the list of subscribers.
+         */
+        this._tickifiedFunctions[cbString] = { tickSubId, wrapperFunc };
+
+        return wrapperFunc;
     }
 
     /**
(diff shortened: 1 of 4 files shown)


src/viewer/scene/scene/Scene.js
```diff
@@ -2562,6 +2567,63 @@ class Scene extends Component {
...
+     * 
+     * This means, the function will be wrapped so:
+     * 
+     * - it runs time-aligned to scene ticks
+     * - it runs maximum once per scene-tick
+     * 
+     * @param {Function} cb The function to tickify
+     * @returns {Function)}
+     */
+    tickify(cb) {
```
xeolabs, line 2581: Maybe an `untickify` method as well? Then it can support components that get destroyed during the lifetime of a Viewer.

tmarti: Please let me finish the PR description 😂

xeolabs:
On my end, I'm also adding some results caching in `PickController`, hashed on the canvas pick coords, to invalidate on each "rendering" event. Will fit OK with this PR.

tmarti:
The PR description is now complete 🎉

PR #1308, opened 2023-12-18
Refactor SceneModel; Add snapping for line and point primitives

xeolabs:


pbbr:
@xeolabs Very nice extension, thanks! We saw that the pick buffer content won't be cleared for some reasons (https://github.com/xeokit/xeokit-sdk/blob/a02fc5abfcdf8b7c1392052f2eb3cc4666d96289/src/viewer/scene/webgl/Renderer.js#L1341C24-L1341C24). This can lead to inconsistent results. Do you have an idea how to fix this?

pbbr:
@xeolabs Furthermore, there are some cases where the returned entity seems to be incorrect (but it's overlapping with the correct one). Do you have an idea why this happens?

https://github.com/xeokit/xeokit-sdk/assets/79220314/41fc748a-406e-4ecb-9929-b6c03b1fc31b

xeolabs:
> @xeolabs Very nice extension, thanks! We saw that the pick buffer content won't be cleared for some reasons (https://github.com/xeokit/xeokit-sdk/blob/a02fc5abfcdf8b7c1392052f2eb3cc4666d96289/src/viewer/scene/webgl/Renderer.js#L1341C24-L1341C24). This can lead to inconsistent results. Do you have an idea how to fix this?

Great - no, I've no idea at this stage what's causing this issue, will need to get a second pair of eyes on this code at some point

xeolabs:
> @xeolabs Furthermore, there are some cases where the returned entity seems to be incorrect (but it's overlapping with the correct one). Do you have an idea why this happens?
> 
>  2024-01-06.130816.mp4

Not sure about this one either at this point - this one also I'll need to get another pair of eyes on this one too.

pbbr:
@xeolabs Thanks for your feedback. An idea to improve the picking is to render and use the edges and vertices for the snap pick entity (https://github.com/xeokit/xeokit-sdk/blob/a02fc5abfcdf8b7c1392052f2eb3cc4666d96289/src/viewer/scene/webgl/Renderer.js#L1462). With this, the picking between normal and snapped result are separated. But this doesn't solve the effect, that the snapped entity could be different from the entity where the mouse pointer is ove. What do you think?

PR #1327, opened 2024-01-16
Removes 404 background in examples by fixing a path

paireks:
Fixes 404 background by providing correct path

Before:
2024-01-16_18h48_43

After:
2024-01-16_18h49_35


xeolabs:
Thanks @paireks !

Issue #1211, opened 2023-11-03
NavCube Text Color


AXONE-IO:
**Is your feature request related to a problem? Please describe.**
We cannot modify the color of the NavCube labels.

**Describe the solution you'd like**
Expose the property text color.

PR #1328, opened 2024-01-16
Custom label color for NavCube

paireks:
- Adding ability to set a label color as described here: https://github.com/xeokit/xeokit-sdk/issues/1211
- Tiny fix of formatting in param for cfg.color (missing quotation mark)

Question if example file (https://github.com/xeokit/xeokit-sdk/blob/master/examples/navigation/NavCubePlugin.html) should be updated to show this feature.

Default color is set as black to don't change existing behaviour.

### DIFF
--- a/examples/navigation/NavCubePlugin_customColors.html
+++ b/examples/navigation/NavCubePlugin_customColors.html
@@ -90,10 +90,14 @@
         topColor: "#5555FF",
         bottomColor: "#fff155",
 
-        // We can also supply a color to highlight NavCuve regions
+        // We can also supply a color to highlight NavCube regions
         // as we hover the pointer over them:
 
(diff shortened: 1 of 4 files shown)


xeolabs:
Looks good to me @paireks - yeah if you could also update the types as @mlankamp suggested, that would be perfect, then I'll merge.

xeolabs:
Also @paireks yes, if you could update this example to show off the custom label colors that would be ideal: https://github.com/xeokit/xeokit-sdk/blob/master/examples/navigation/NavCubePlugin_customColors.html

paireks:
Here it is.
Tiny question: when xeokit-sdk.min.es.js is updated?

xeolabs:
> Here it is. Tiny question: when xeokit-sdk.min.es.js is updated?

I build that locally in the master branch, after updating the npm version in package.json etc.

PR #1332, opened 2024-01-17
[Fix] Fixed links to assets, removed link to roboto condensed

mlankamp:
Some example files had wrong links to assets, for example: https://xeokit.github.io/xeokit-sdk/examples/navigation/#TreeViewPlugin_Containment

All the index.html files had a stylesheet link to http://fonts.googleapis.com, since the example site is loaded over https the file was not loaded (and not needed). So removed from all files


xeolabs:
Thanks @mlankamp !

Issue #1325, opened 2024-01-12
[FEATURE]:Enable external render functions for treeview


mlankamp:
In the TreeViewPlugin move the functions:
- _createNodeElement
- _expandSwitchElement
- _collapseNode
- _collapseSwitchElement

to the constructor and allow these functions to be overwritten via the constructor configuration.

This allows users to create their own styling. 

Alternative solution:
Create a render service, which can be passed via the constructor (same as the localeService).

mlankamp:
I created a PR to support this scenario #1329

xeolabs:
Looks good, thanks @mlankamp. Could we please also get an HTML example page to demo it? Maybe just base it off this one: https://github.com/xeokit/xeokit-sdk/blob/master/examples/navigation/TreeViewPlugin_Containment.html

mlankamp:
Yes, I will create an example! Will include the x-rayed style as well.

mlankamp:
Done, see PR #1331

mlankamp:
@xeolabs is it possible to create a new beta? Will upgrade my typescript based solution and can check if the typings are ok

xeolabs:
@mlankamp yip, done - 2.5.2-beta-5

mlankamp:
Nice, with the latest merge the example works: https://xeokit.github.io/xeokit-sdk/examples/navigation/#TreeViewPlugin_options_renderService

xeolabs:
Could it be extended to create IFC element icons on the tree nodes, do you think? Maybe extend the callback to take a MetaObject, then add class names to node elements to style them with the corresponding icon?

mlankamp:
That should be possible, will try tomorrow

mlankamp:
The MetaObject contains the type, which I can use to render an icon in the treeview. I came up with this example:
![image](https://github.com/xeokit/xeokit-sdk/assets/8513556/c4213f05-969c-4f1d-86ad-149a18bf93d5)

We could a helper class, which maps a ifc type to a specific icon (font-awesome and material design icons). Or even better, is there a defined icon set?

mlankamp:
Found this: https://aecgeeks.github.io/ifc-icons/ (source: https://github.com/AECgeeks/ifc-icons)

xeolabs:
> Found this: https://aecgeeks.github.io/ifc-icons/ (source: https://github.com/AECgeeks/ifc-icons)

Yip that's the one I had in mind!

[...]

PR #1334, opened 2024-01-18
[EXAMPLE] Add icons for different types of nodes in tree view

mlankamp:
Based on https://aecgeeks.github.io/ifc-icons/, altered to use material design icons

see #1325


xeolabs:
Nice - let me know when it's ready and I'll merge

Issue #1336, opened 2024-01-18
Excess meshes with SceneModel.createMesh() causes problems 


xeolabs:
If we create meshes with SceneModel.createMesh() but don't actually use those meshes with SceneModel.createEntity(), then we end up with internal counters (numVisible, numHighlighted etc) becoming incorrect.

This is because certain internal counts are initialised by SceneModel.createMesh(), independently of whether SceneModel.createEntity() is ever called.

Easiest fix is to make SceneModelfinalize() fail if there are any meshes that are not used by any entities.

PR #1338, opened 2024-01-18
Throws when one try to create a scene with an unused mesh during finalize.

paireks:
I tried to implement easiest fix of this issue: https://github.com/xeokit/xeokit-sdk/issues/1336

Please have a look at that proposal and give me hints if I got something wrong.
When someone tries to create a mesh that is later not used -> error should be thrown during finalize.
It's a quite big change of existing behaviour.

### DIFF
--- a/src/viewer/scene/model/SceneModel.js
+++ b/src/viewer/scene/model/SceneModel.js
@@ -1146,8 +1146,6 @@ export class SceneModel extends Component {
         this._vboBatchingLayers = {};
         this._dtxLayers = {};
 
-        this._meshList = [];
-
         this.layerList = []; // For GL state efficiency when drawing, InstancingLayers are in first part, BatchingLayers are in second
         this._entityList = [];
 
@@ -1159,7 +1157,8 @@ export class SceneModel extends Component {
         this._meshes = {};
         this._entities = {};
 
-        this._scheduledMeshes = {}
+        this._scheduledMeshes = {};
+        this._meshesCfgsBeforeMeshCreation = {};
 
         /** @private **/
         this.renderFlags = new RenderFlags();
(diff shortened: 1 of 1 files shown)


xeolabs:
Thanks - let me think about this one for a bit though - throwing an exception is a bit brutal. It might actually be a better to create dummy entities for those meshes, with object:false.

paireks:
Sure, but should we create these dummy entities during finalize?

paireks:
Or maybe there is some other way to inform the user inside viewer or console about the issue, without throwing an exception?

xeolabs:
> Or maybe there is some other way to inform the user inside viewer or console about the issue, without throwing an exception?

Perhaps both creating the entities in finalize() and at the same time logging a warning to the console.

xeolabs:
Actually, it may be practical to defer the creation of meshes until their entities are created - see here: https://github.com/xeokit/xeokit-sdk/blob/master/src/viewer/scene/model/SceneModel.js#L3050

The mesh parameters could be cached, and then SceneModel._createMesh() called to create the mesh within SceneModel.createEntity() only once the Entity refers to it.

xeolabs:
In that code, we're creating the mesh immediately at the end of SceneModel.createMesh(), but that `cfg` passed to `_createMesh` could be stored in a map against the mesh ID, then SceneModel.createEnity() could get that from the map and call  `_createMesh` with it.

paireks:
If I understand it correctly, you'd like to postpone mesh creation, which makes sense to me, but it would also be probably a breaking change, as SceneModel.createMesh() already returns an instantly created mesh.

xeolabs:
> If I understand it correctly, you'd like to postpone mesh creation, which makes sense to me, but it would also be probably a breaking change, as SceneModel.createMesh() already returns an instantly created mesh.

Yes, and actually this API change would be OK, because nothing outside of SceneModel ever uses those returned SceneMeshes - they are only used by SceneObjects. I just made createMesh() return a SceneMesh to make testing easier, and for symmetry. It would be fine to make that method return nothing.

paireks:
Ok, so here is mesh creation deferring. I attached comment for each line in deferring part as it is a bit more tricky, but I can remove it any time.
Let me know if I got something wrong.
Also I return true/false for createMesh(), but can return nothing if it's preferred.

paireks:
I will update types part after this part will be accepted.

xeolabs:
Looks good to me @paireks, thanks

xeolabs:
Published in 2.5.2-beta-7 🙏

Issue #1065, opened 2023-05-30
Measurement label layout to avoid clutter


xeolabs:
![Screenshot from 2023-05-30 13-21-00](https://github.com/xeokit/xeokit-sdk/assets/83100/049b9d80-2352-415f-a6be-7ecb2d2b5f07)
![Screenshot from 2023-05-30 13-20-50](https://github.com/xeokit/xeokit-sdk/assets/83100/47b2bd25-4042-45af-9ad1-d51c7f2a3b24)

PR #1365, opened 2024-02-10
[FEATURE] Adding ability to show labels one below the other for measurement plugin

paireks:
Related to: https://github.com/xeokit/xeokit-sdk/issues/1065

2024-02-10_13h37_22

2024-02-10_17h48_33

Few notes:
- update was made in a way to not break existing behaviour. User needs to set labelsOnWires to false to display it in a new way.
- it was made on purpose to display all the time X, Y, Z once we choose to show one label below another one. The other option would be to try to hide the ones that are zero's (same as it is with labelsOnWires), but after experimenting a while with it I choose to stick to showing X, Y, Z no matter of the value. It was a subjective decision, so it can be questionable.
- there are 2 examples updated: 1 new (copy of existing one, but with changed value) and 1 updated (updated one had 2 existing measurements, so I set 1 with new settings).
- d.ts files will be updated once this part will be accepted

### DIFF
--- a/examples/measurement/distance_createWithMouse_labelsNotOnWires.html
+++ b/examples/measurement/distance_createWithMouse_labelsNotOnWires.html
@@ -85,10 +85,10 @@
 <canvas id="myCanvas"></canvas>
 <div class="slideout-sidebar">
     <img class="info-icon" src="../../assets/images/measure_distance_icon.png"/>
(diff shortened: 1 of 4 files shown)


xeolabs:
Looks good! Thanks @paireks

PR #1382, opened 2024-02-26
[EXAMPLE] Adding example of controlling the position on the CameraPath by slider

paireks:
This example is a byproduct of me trying to figure out how to control the position on the path. Please decide if it's worth adding.

https://github.com/xeokit/xeokit-sdk/assets/47977819/198efe44-d45c-4e18-b560-038289186aef


xeolabs:
This is cool - one tweak - could position the slider along the full width of the canvas? Seems the side bar makes it really short.

paireks:
Just to make sure, you have in mind something like this in the case 2 here in the picture?

image

paireks:
https://github.com/xeokit/xeokit-sdk/assets/47977819/d804d705-433f-44fc-aced-0317e754ed14

Like this?

xeolabs:
Yeah perfect. If you can make it resize when the info panel on the right is opened, like to always scale to the canvas size, that would be ideal.

PR #1390, opened 2024-02-29
[EXAMPLE] Section path example

paireks:
Example of cutting the model along the section path.
Additionally there is a tiny refactor on previous examples, to make sure there is a casting from slider.value to Number.

https://github.com/xeokit/xeokit-sdk/assets/47977819/6b3e6fbb-6477-44fe-8d36-b8a8edd4013d


xeolabs:
Nice, that's a cool idea!

Issue #1391, opened 2024-03-01
Fix code scanning alert - Unsafe dynamic method access


drashevski:
Tracking issue for:
- [ ] https://github.com/xeokit/xeokit-sdk/security/code-scanning/5

PR #1399, opened 2024-03-05
Refactor of z-worker.js (WIP)

paireks:
I'm trying to fix issue mentioned here: https://github.com/xeokit/xeokit-sdk/issues/1391

But I don't know if and how refactor made make it more reliable. I guess one way to check that would be to run the scan again after the change, but I won't be suprised if scan itself wouldn't be 100% reliable, maybe it won't see the problem anymore, but the problem will still occur. Best test would be to actually trying to break it and see if we can break it before and after the refactor. But I'll need some help how to write such test. Please let me know Lindsay if you have some ideas how to deal with such fixes.


paireks:
I'm closing it for now, once we decide to work on it it can be reopened

Issue #1403, opened 2024-03-07
Need to load different MTL loading the same .obj with ObjLoaderPlugin


erikwski:
**Is your feature request related to a problem? Please describe.**
I load an xkt model and need to append the same obj model multiple time. 
If i do that and my .mtl file use the keyword newmtl for create new Material, i will get an error for same material id on PhongMaterial.

**Describe the solution you'd like**
I would like to doesn't need to make manual changes loading different .mtl or use diffent id. Maybe the id of the Material can become `id=${newMtlId}-${objId}`.

**Describe alternatives you've considered**
Load different .mtl, one using newmtl, and the second time use the file with usemtl and retrieve PhongMaterial instance and set to it

xeolabs:
If I understand correctly:

- you load an XKT file
- then load the same OBJ file multiple times
- find there are ID clashes between the PhongMaterials created for the OBJ models

Unfortunately we don't currently support this use case, where we don't support loading multiple copies of the same OBJ model.

If anyone wants to extend OBJLoaderPlugin to enable the PhongMaterials to get non-clashing IDs for this case, we'd consider a PR.

I hope this helps..

PR #1404, opened 2024-03-07
Fixed duplicate loading of .obj model causing material ID errors

erikwski:
I'm excited to make my first contribution to the project!

This pull request addresses the issue described in the closed issue: https://github.com/xeokit/xeokit-sdk/issues/1403, where loading an .obj model could result in the creation of unnecessary "Default" materials or duplicate materials.

I've implemented a solution based on approaches used in my previous projects, which effectively prevents these issues.

### Key changes:
- Don't create a Default material everytime that parse a MTL file
- The Material id is generated using the node id + "_" + material.id
- If the material of the same type of node try to be created again, just link the instance to the new component setted in _ownedComponents attributes

### Testing:

I've thoroughly tested this solution in my local environment to ensure it resolves the issue without introducing any regressions.
I welcome any feedback or suggestions you may have.

Thank you for your time!

### DIFF
--- a/src/plugins/OBJLoaderPlugin/OBJSceneGraphLoader.js
+++ b/src/plugins/OBJLoaderPlugin/OBJSceneGraphLoader.js
@@ -650,7 +650,14 @@ var parseMTL = (function () {
     }
 
     function createMaterial(modelNode, materialCfg) {
-        new PhongMaterial(modelNode, materialCfg);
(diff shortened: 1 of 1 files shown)


erikwski:
I closed that because in some case the change of the id create an error and the .obj cannot find the Material

erikwski:
I'm going to fix that and open new PR

PR #1434, opened 2024-04-10
FIX Viewer.js breaking import

Kurtil:
This PR reverts this import to its previous value. The current import is breaking xeokit if used as an npm dependency.

### DIFF
--- a/src/viewer/Viewer.js
+++ b/src/viewer/Viewer.js
@@ -3,7 +3,7 @@ import {CameraFlightAnimation} from "./scene/camera/CameraFlightAnimation.js";
 import {CameraControl} from "./scene/CameraControl/CameraControl.js";
 import {MetaScene} from "./metadata/MetaScene.js";
 import {LocaleService} from "./localization/LocaleService.js";
-import html2canvas from '../../node_modules/html2canvas/dist/html2canvas.esm.js';
+import html2canvas from 'html2canvas/dist/html2canvas.esm.js';
 
 /**
  * The 3D Viewer at the heart of the xeokit SDK.


xeolabs:
Published in https://github.com/xeokit/xeokit-sdk/releases/tag/v2.6.0-beta-12

paireks:
@Kurtil Hello, thank you for correcting me :) Question is if there is a third option, because everytime I want to develop by importing Viewer.js like this: `import {Viewer} from "../../src/viewer/Viewer.js"`
I have an error:
2024-04-21_12h29_14
And I have to manually switch it to this other type of import inside Viewer.js file, and later remove it. Do you have an idea how to solve this, so both this and the npm dependency will work?

Kurtil:
@paireks I had the same issue. This issue is due to the fact that this import is not a valid ESM javascript import, it is an import that must be handled first by a bundler.

I imagine you serve directly your file using a local web server without bundling it like I do (seeing that we have this same error). In this case, you can comment the html2canvas import line if you do not use the corresponding lines like getSnapshot if I remember correctly, or a nice solution I use now is to use import map in the `head` of the `html` file you serve:

```html
<script type="importmap">
  {
    "imports": {
      "html2canvas/dist/html2canvas.esm.js": "https://cdn.jsdelivr.net/npm/html2canvas@1.4.1/dist/html2canvas.esm.js"
    }
  }
</script>
```

xeolabs:
Would it make sense to extend the build process to copy html2canvas from node_modules into the ./src directory? Then it becomes part of the source tree.

Kurtil:
Resolve the path using a bundler is more "js friendly" I assume...

PR #1445, opened 2024-04-18
Feature - Add support for Pointer Lock Web API

Kurtil:
I'm developing a product in which the pointer is locked during the camera update when the first-person view is activated.

pointerLocked

On the GIF, the pointer does not move because it is locked. It lets you update the camera without moving the pointer and risking losing it in a corner of the screen.

To lock the pointer, I use the [Pointer Lock WebAPI](https://developer.mozilla.org/en-US/docs/Web/API/Pointer_Lock_API).

```js
viewer.scene.canvas.canvas.requestPointerLock();
```

**Issue**: The mouse pan/rotate/dolly handler [uses the pointer canvas position to update the camera control state](https://github.com/xeokit/xeokit-sdk/blob/master/src/viewer/scene/CameraControl/lib/handlers/MousePanRotateDollyHandler.js#L175-L176) but when the pointer is locked, the pointer does not move.

**Solution** : This PR check if the pointer is locked with `document.pointerLockElement` and if so, the `MouseEvent<"mousemove">` `movementX` and `movementY` properties are used instead of the real pointer canvas position.

### DIFF
--- a/src/viewer/scene/CameraControl/lib/handlers/MousePanRotateDollyHandler.js
+++ b/src/viewer/scene/CameraControl/lib/handlers/MousePanRotateDollyHandler.js
@@ -156,7 +156,7 @@ class MousePanRotateDollyHandler {
             }
         });
 
-        document.addEventListener("mousemove", this._documentMouseMoveHandler = () => {
+        document.addEventListener("mousemove", this._documentMouseMoveHandler = (e) => {
 
             if (!(configs.active && configs.pointerEnabled)) {
                 return;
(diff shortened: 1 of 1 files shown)


xeolabs:
Thanks @Kurtil - could you please also add an example of how to use pointer lock? Would be great for our testing/QA - maybe copy this example: https://github.com/xeokit/xeokit-sdk/blob/master/examples/navigation/CameraControl_firstPerson_MAP.html

Kurtil:
Ok @xeolabs , I'll add an example.

Kurtil:
Is [this](https://github.com/xeokit/xeokit-sdk/pull/1455) ok ? :)

xeolabs:
Yeah that great, thanks @Kurtil

xeolabs:
I just moved the example into it's own page, just to highlight the feature a bit. It's a cool feature! 
https://github.com/xeokit/xeokit-sdk/pull/1456

Issue #1461, opened 2024-04-22
[FIX] Ensure no interference between ContextMenus on measurements and Canvas


xeolabs:
When we have a `ContextMenu` attached to the canvas, and another `ContextMenu` attached to a measurement label, then we need to ensure that a right-click on the measurement label opens the `ConextMenu` only on the label, and not on the canvas.

Create an example to demonstrate how to set up `ContextMenus` for this case.

PR #1462, opened 2024-04-22
[FIX] Fix interference between ContextMenus on measurements and Canvas

xeolabs:
Addresses  #1461

- Create an example that demonstrates how to have context menus on measurements AND canvas, such that there is not interference between the right-click input used to open each menu.
- Ensure that events are properly consumed by measurement HTML elements.


xeolabs:
Fixed in xeokit-sdk 2.6.3: https://github.com/xeokit/xeokit-sdk/releases/tag/v2.6.3

PR #1510, opened 2024-05-28
[FEATURE] DotBIMLoaderPlugin

xeolabs:
This PR adds the `DotBIMLoaderPlugin` class, which enables us to to load `.bim` models into a xeokit `Viewer`. 

When loading a `.bim` model, the  `DotBIMLoader` does the following:

* Creates an `Entity` representing each `.bim` model it loads, which will have `Entity#isModel` set ````true````
and will be registered by `Entity#id` in `Scene#models`.
* Creates an `Entity` for each object within the `.bim` model. Those `Entities` will have `Entity#isObject`
set ````true```` and will be registered by `Entity#id` in `Scene#objects`.
* When loading, can set the World-space position, scale and rotation of each model within World space,
along with initial properties for all the model's `Entity`s.
* Allows to mask which IFC types we want to load.
* Allows to configure initial viewer state for specified IFC types (color, visibility, selection, highlighted, X-rayed, pickable, etc).

# Usage

In the example below we'll load a house model from a .bim file.

This will create a bunch of `Entity`s that represents the model and its objects, along with
a `MetaModel` and `MetaObject`s that hold their metadata.

Screenshot from 2024-05-28 22-32-05

````javascript
import {Viewer, DotBIMLoaderPlugin} from "xeokit-sdk.es.js";

//------------------------------------------------------------------------------------------------------------------
// 1. Create a Viewer,
// 2. Arrange the camera,
// 3. Tweak the selection material (tone it down a bit)
//------------------------------------------------------------------------------------------------------------------

// 1
const viewer = new Viewer({
     canvasId: "myCanvas",
     transparent: true
});

// 2
viewer.camera.orbitPitch(20);
viewer.camera.orbitYaw(-45);

// 3
viewer.scene.selectedMaterial.fillAlpha = 0.1;

//------------------------------------------------------------------------------------------------------------------
// 1. Create a .bim loader plugin,
// 2. Load a .bim building model, emphasizing the edges to make it look nicer
//------------------------------------------------------------------------------------------------------------------

// 1
const dotBIMLoader = new DotBIMLoaderPlugin(viewer);

// 2
var model = dotBIMLoader.load({                                    // Returns an Entity that represents the model
     id: "myModel",
     src: "../../assets/models/dotbim/House.bim",
     edges: true
});

// Find the model Entity by ID
model = viewer.scene.models["myModel"];

// Destroy the model
model.destroy();
````

## Transforming the Model

We have the option to rotate, scale and translate each  *````.bim````* model as we load it.

This lets us load multiple models, or even multiple copies of the same model, and position them apart from each other.

In the example below, we'll rotate our model 90 degrees about its local X-axis, then
translate it 100 units along its X axis.

````javascript
const model = dotBIMLoader.load({
     src: "../../assets/models/dotbim/House.bim",
     rotation: [90,0,0],
     position: [100, 0, 0]
});
````

## Including and excluding IFC types

We can also load only those objects that have the specified IFC types. In the example below, we'll load only the
objects that represent walls.

````javascript
const model = dotBIMLoader.load({
    id: "myModel",
     src: "../../assets/models/dotbim/House.bim",
     includeTypes: ["IfcWallStandardCase"]
});
````

We can also load only those objects that **don't** have the specified IFC types. In the example below, we'll load only the
objects that do not represent empty space.

````javascript
const model = dotBIMLoader.load({
    id: "myModel",
     src: "../../assets/models/dotbim/House.bim",
     excludeTypes: ["IfcSpace"]
});
````

## Configuring initial IFC object appearances

We can specify the custom initial appearance of loaded objects according to their IFC types.

This is useful for things like:

* setting the colors to our objects according to their IFC types,
* automatically hiding ````IfcSpace```` objects, and
* ensuring that ````IfcWindow```` objects are always transparent.
<br>
In the example below, we'll load a model, while configuring ````IfcSpace```` elements to be always initially invisible,
and ````IfcWindow```` types to be always translucent blue.

````javascript
const myObjectDefaults = {

     IfcSpace: {
         visible: false
     },
     IfcWindow: {
         colorize: [0.337255, 0.303922, 0.870588], // Blue
         opacity: 0.3
     },

     //...

     DEFAULT: {
         colorize: [0.5, 0.5, 0.5]
     }
};

const model4 = dotBIMLoader.load({
     id: "myModel4",
     src: "House.bim",
     objectDefaults: myObjectDefaults // Use our custom initial default states for object Entities
});
````

When we don't customize the appearance of IFC types, as just above, then IfcSpace elements tend to obscure other
elements, which can be confusing.

It's often helpful to make IfcSpaces transparent and unpickable, like this:

````javascript
const dotBIMLoader = new DotBIMLoaderPlugin(viewer, {
   objectDefaults: {
       IfcSpace: {
           pickable: false,
           opacity: 0.2
       }
   }
});
````

Alternatively, we could just make IfcSpaces invisible, which also makes them unpickable:

````javascript
const dotBIMLoader = new DotBIMLoaderPlugin(viewer, {
   objectDefaults: {
       IfcSpace: {
           visible: false
       }
   }
});
````

## Configuring a custom data source

By default, DotBIMLoaderPlugin will load *````.bim````* files and metadata JSON over HTTP.

In the example below, we'll customize the way DotBIMLoaderPlugin loads the files by configuring it with our own data source
object. For simplicity, our custom data source example also uses HTTP, using a couple of xeokit utility functions.

````javascript
import {utils} from "xeokit-sdk.es.js";

class MyDataSource {

     constructor() {
     }

     // Gets the contents of the given .bim file in a JSON object
     getDotBIM(src, ok, error) {
         utils.loadJSON(dotBIMSrc,
            (json) => {
                ok(json);
            },
            function (errMsg) {
                error(errMsg);
            });
     }
}

const dotBIMLoader2 = new DotBIMLoaderPlugin(viewer, {
      dataSource: new MyDataSource()
});

const model5 = dotBIMLoader2.load({
     id: "myModel5",
     src: "House.bim"
});
````


xeolabs:
/ping @paireks 

One little hack I made in DotBIMLoaderPlugin is that I autogenerate some top-level IFC elements in the metadata for each .bim model - see here: https://github.com/xeokit/xeokit-sdk/blob/master/src/plugins/DotBIMLoaderPlugin/DotBIMLoaderPlugin.js#L358

I generate the IfcProject, IfcSite, IfcBuilding elements, plus a default IfcBuildingStory, and I make all the .bim elements children of that IfcBuildingStory.

That's a pretty rough hack for sure, using just one storey. For now, it just allows everything to appear in a TreeViewPlugin's tree view for its "storeys" hierarchy mode, in addition to its "containment" and "types" modes, and also allows the elements to appear in a 2D plan view generated by StoreyViewsPlugin.

Maybe we can think of a better way to organize the elements into storeys.. that's a hack that we can use just to get started I think.

paireks:
@paireks ok, now I see why there is an IfcProject at the top of a tree view :)

xeolabs:
Hi @paireks when an element has an `info` that contains `IFC_Pset_` properties, should the DotBIMLoader create a separate xeokit `PropertySet` for that element, containing those `IFC_Pset_`  properties?

Or perhaps the `DotBIMLoader` should try to merge common properties into shared `ProperySets`, to save memory?

paireks:
Hey @xeolabs, if I understand correctly: I'm okay with having separate PropertySet for each element without merging it. 
And only once we will find some files that cause the memory issue we could try to merge it, what do you think?

xeolabs:
Hey @paireks yip sounds good to me.

xeolabs:
Battle station fully operational, you may fire when ready.

paireks:
@xeolabs :)

Do you think it could make sense to prepare also a large (in terms of kilometers long) .bim file to show advantage of double-precision coordinates? If so: how big the file should be to see it's working?

xeolabs:
> @xeolabs :)
> 
> Do you think it could make sense to prepare also a large (in terms of kilometers long) .bim file to show advantage of double-precision coordinates? If so: how big the file should be to see it's working?

Yes that's a good idea. Although we'd need the objects to not be really long, ie. thousands of units long, where they would need splitting to fit into the tiles. So no super long railway lines etc.

 If we can have objects each of, say no more than a couple of hundred meters each long, that would be ideal. In xeokit, 1.0 coordinates units is one meter BTW.
 
 Another option is to provide a building that's not particularly long, but positioned far from the origin, so that it has huge coordinates.

PR #1514, opened 2024-05-29
[EXAMPLE] Add non-ifc->.bim examples

paireks:
Hello,

first of all: I'm super excited about dotbimLoader, and I will try to use it myself. Thank you!

In this PR I've added 2 .bim files examples that don't come from IFC -> .bim conversion. I wanted to add these, because IFC -> .bim conversion is just one specific option to creat .bim files, however idea of format was to be used it as a direct export from specific types of software, without the need of any conversion. All existing connectors and exports get meshes + data directly from modelling software to dotbim.

Feel free to decide if it's worth adding, and also if you will have some other idea of a file that could be added there, or some corner-case you'd like to check: let me know, I will prepare such file.

Thank you.


xeolabs:
Looks good to me, thanks!

PR #1517, opened 2024-06-01
[EXAMPLE] Change teapot example to test also double-precision precision

paireks:
I modified teapot-benchmark example so it maybe can present a bit more xeokit capabilities:
- there should be 10000+ teapots
- they are populated on the ~1500 meters long space
- I cannot load it in any other web viewer

2024-06-01_21h04_12

Let me know what you think. And also: if you will need and have an idea of some benchmark file - let me know, I will try to create it.


xeolabs:
That's excellent, thanks!

PR #1529, opened 2024-06-11
[FEATURE] Styled lines

paireks:
Hello,

here is an additional method to create styled lines in a simple way
2024-06-11_21h36_05

PR already includes example and new .d.ts file.

I was adding this having in mind addition of grid lines in the future, it might be helpful for it.

Looking forward to your feedback.

### DIFF
--- a/examples/scenegraph/buildLineGeometry.html
+++ b/examples/scenegraph/buildLineGeometry.html
@@ -70,7 +70,7 @@
     viewer.camera.up = [0, 1, 0];
 
     //------------------------------------------------------------------------------------------------------------------
-    // Create a mesh with simple line shape
+    // Create a mesh with simple 2d line shape
     //------------------------------------------------------------------------------------------------------------------
 
     new Mesh(viewer.scene, {
@@ -84,7 +84,7 @@
     });
 
     //------------------------------------------------------------------------------------------------------------------
-    // Create a mesh with simple line shape with black color
+    // Create a mesh with simple 2d line shape with black color
     //------------------------------------------------------------------------------------------------------------------
 
     new Mesh(viewer.scene, {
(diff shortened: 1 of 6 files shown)


xeolabs:
Cool @paireks I just went ahead and merged this because it looks good. Cool benefit of doing the dashes on the actual geometry is that you'll get the perspective reduction of the dashes themselves. 

Thinking out loud: Perhaps we could think of adding this to all the generator functions, that generate line geometry. If so, it might be a good idea to make the dashing function a parameter that can be passed into the generator functions as a parameter, just for open/closed principle goodness. Or should the pattern parameter just be the API's standard way to specify dashes? Maybe both?

I wonder if there is any case for using such a function within the current way we generate edges, that we currently use for geometry. Maybe `SceneModel.createGeometry()` and `SceneModel.createMesh()` could have an `edgesPattern` and/or `edgesPatterner()` parameter?

paireks:
> all the generator functions, that generate line geometry

Just to make sure: do you think about other geometry builders only (such as buildGridGeometry, buildPolylineGeometry etc.) or some other places?

xeolabs:
> > all the generator functions, that generate line geometry
> 
> Just to make sure: do you think about other geometry builders only (such as buildGridGeometry, buildPolylineGeometry etc.) or some other places?

I mean those, yip

paireks:
I will have to think about it for a while and experiment with existing paths functions and will come back to you. Cause I think I can't reuse this logic for dashing creation for a single line to all other types of curves, I have to check if each curve type should have their own separate pattern creation logic, or maybe developing one for Path.js is possible. Or maybe we could limit it to straight lines only. Generally I have to think about it a bit longer :)

PR #1532, opened 2024-06-11
[FEATURE] Extend pattern to end point

paireks:
Here I just wanted to add an option to extend the last segment to end point (this option is turned on for blue lines below). It could be helpful if someone wants to see the real length / start and end point of the line, but can deal with the fact pattern at the end is slightly different. It is a usual practice in different programs, but handled in a various different ways.

2024-06-11_23h44_16

### DIFF
--- a/examples/scenegraph/buildLineGeometry.html
+++ b/examples/scenegraph/buildLineGeometry.html
@@ -112,17 +112,49 @@
     });
 
     //------------------------------------------------------------------------------------------------------------------
-    // Create a mesh with simple 2d line shape with black color and more complex pattern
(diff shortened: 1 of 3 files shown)


xeolabs:
Cool, makes sense. Another thought: this dashing is similar to [CanvasRenderingContext2D: setLineDash()](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/setLineDash), and therefore quite conventional, so I think it's worth actually building it into xeokit via a `pattern` parameter option for wherever lines are generated, rather than use a callback function, like I was thinking earlier.

PR #1536, opened 2024-06-14
[FIX] Making optional control of precision of annotation occluding

paireks:
This PR includes:
- revert of offset in a shader to make existing behaviour unchanged
- adding parameter for the user to control the offset, so he can decide if he want to make it more precise or not
- 2 new examples with usage of this additional parameter, for which changing of the offset parameter makes occluding better
- update of d.ts file

Looking forward for review.

### DIFF
--- a/examples/annotations/annotations_createWithMouse_orthoWithZMarkerOffset.html
+++ b/examples/annotations/annotations_createWithMouse_orthoWithZMarkerOffset.html
@@ -91,7 +91,7 @@
 <div class="slideout-sidebar">
     <img class="info-icon" src="../../assets/images/annotation_icon.png"/>
     <h1>AnnotationsPlugin</h1>
-    <h2>Click or tap the model to create annotations, control OcclusionTester.js precision</h2>
+    <h2>Click or tap the model to create annotations, control occlusion precision with markerZOffset</h2>
     <h3>Components Used</h3>
     <ul>
         <li>
--- a/examples/annotations/annotations_createWithMouse_orthoWithZMarkerOffset.html
+++ b/examples/annotations/annotations_createWithMouse_orthoWithZMarkerOffset.html
@@ -91,7 +91,7 @@
 <div class="slideout-sidebar">
     <img class="info-icon" src="../../assets/images/annotation_icon.png"/>
     <h1>AnnotationsPlugin</h1>
-    <h2>Click or tap the model to create annotations</h2>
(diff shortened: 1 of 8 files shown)


examples/annotations/annotations_createWithMouse_orthoWithZMarkerOffset.html
```diff
@@ -0,0 +1,223 @@
...
+</head>
+
+<body>
+<input type="checkbox" id="info-button"/>
+<label for="info-button" class="info-button"><i class="far fa-3x fa-question-circle"></i></label>
+<canvas id="myCanvas"></canvas>
+<div class="slideout-sidebar">
+    <img class="info-icon" src="../../assets/images/annotation_icon.png"/>
+    <h1>AnnotationsPlugin</h1>
+    <h2>Click or tap the model to create annotations, control OcclusionTester.js precision</h2>
```
xeolabs, line 94: Could we rename this to "Click or tap the model to create annotations, with configured marketZOffset"
So as not to mention OcclusionTester.js, which is private code

examples/annotations/annotations_createWithMouse_orthoWithZMarkerOffsetZero.html
```diff
@@ -0,0 +1,229 @@
...
+</head>
+
+<body>
+<input type="checkbox" id="info-button"/>
+<label for="info-button" class="info-button"><i class="far fa-3x fa-question-circle"></i></label>
+<canvas id="myCanvas"></canvas>
+<div class="slideout-sidebar">
+    <img class="info-icon" src="../../assets/images/annotation_icon.png"/>
+    <h1>AnnotationsPlugin</h1>
+    <h2>Click or tap the model to create annotations, control OcclusionTester.js precision</h2>
```
xeolabs, line 94: Same here

xeolabs (changes requested):

examples/annotations/annotations_createWithMouse_orthoWithZMarkerOffset.html
paireks: Done

examples/annotations/annotations_createWithMouse_orthoWithZMarkerOffsetZero.html
paireks: Done

PR #1545, opened 2024-06-18
XEOK-36 Edit Measurement

MichalDybizbanskiCreoox:


src/plugins/AngleMeasurementsPlugin/index.js
```diff
@@ -1,4 +1,61 @@
...
+     * Edits {@link AngleMeasurement} with mouse and/or touch input.
+     *
+     * @param {AngleMeasurement} [measurement] The AngleMeasurement to edit.
+     * @param [cfg] Configuration
+     * @param {PointerLens} [cfg.pointerLens] A PointerLens to use to provide a magnified view of the cursor.
+     * @param {boolean} [cfg.snapping] Whether to enable snap-to-vertex and snap-to-edge.
+     * @param {boolean} [handleMouseEvents] Whether to hangle mouse input.
+     * @param {boolean} [handleTouchEvents] Whether to hangle touch input.
+     */
+    constructor(measurement, cfg, handleMouseEvents, handleTouchEvents) {
```
xeolabs, line 21: Could we please put the handleMouseEvents and handleTouchEvents into cfg

src/plugins/DistanceMeasurementsPlugin/index.js
```diff
@@ -1,4 +1,61 @@
...
+     * Edits {@link DistanceMeasurement} with mouse and/or touch input.
+     *
+     * @param {DistanceMeasurement} [measurement] The DistanceMeasurement to edit.
+     * @param [cfg] Configuration
+     * @param {PointerLens} [cfg.pointerLens] A PointerLens to use to provide a magnified view of the cursor.
+     * @param {boolean} [cfg.snapping] Whether to enable snap-to-vertex and snap-to-edge.
+     * @param {boolean} [handleMouseEvents] Whether to hangle mouse input.
+     * @param {boolean} [handleTouchEvents] Whether to hangle touch input.
+     */
+    constructor(measurement, cfg, handleMouseEvents, handleTouchEvents) {
```
xeolabs, line 21: Could we please put the handleMouseEvents and handleTouchEvents into cfg

src/plugins/lib/ui/index.js
```diff
@@ -0,0 +1,222 @@
...
+import {Marker} from "../../../viewer/scene/marker/Marker.js";
+
+export function transformToNode(from, to, vec) {
+    const fromRec = from.getBoundingClientRect();
+    const toRec = to.getBoundingClientRect();
+    vec[0] += fromRec.left - toRec.left;
+    vec[1] += fromRec.top  - toRec.top;
+};
+
+export function draggableDot3D(handleMouseEvents, handleTouchEvents, viewer, worldPos, color, ray2WorldPos, onStart, onMove, onEnd) {
```
xeolabs, line 12: Could we please rename to createDraggableDot3D and use a parameter object (we try to avoid long argument lists just for style)

src/plugins/lib/ui/index.js
```diff
@@ -0,0 +1,222 @@
...
+            currentDrag && currentDrag.cleanup();
+            scene.camera.off(onViewMatrix);
+            scene.camera.off(onProjMatrix);
+            marker.destroy();
+            dot.destroy();
+        }
+    };
+};
+
+export function activateDraggableDots(viewer, handleMouseEvents, handleTouchEvents, snapping, pointerLens, color, markers, onEdit) {
```
xeolabs, line 154: Could we please use a parameter object here as well

xeolabs (changes requested):

src/plugins/AngleMeasurementsPlugin/index.js
MichalDybizbanskiCreoox: Please note the `AngleMeasurementEditControl` is a "private" class, so its arguments are not at the moment part of the public API.
The public classes are `AngleMeasurementEditMouseControl` and `AngleMeasurementEditTouchControl`.
If `handleMouseEvents` and `handleTouchEvents` were to be properties of the `cfg` object, how should these properties be handled if already set inside the public classes' `cfg` argument.

src/plugins/DistanceMeasurementsPlugin/index.js
MichalDybizbanskiCreoox: Please note the `DistanceMeasurementEditControl` is a "private" class, so its arguments are not at the moment part of the public API.
The public classes are `DistanceMeasurementEditMouseControl` and `DistanceMeasurementEditTouchControl`.
If `handleMouseEvents` and `handleTouchEvents` were to be properties of the `cfg` object, how should these properties be handled if already set inside the public classes' `cfg` argument.

xeolabs: Right, keeping handleMouseEvents and handleTouchEvents separate makes sense then

src/plugins/AngleMeasurementsPlugin/index.js
xeolabs: Sure, keep this as is, makes sense

src/plugins/lib/ui/index.js
MichalDybizbanskiCreoox: Implemented by https://github.com/xeokit/xeokit-sdk/commit/72afa7852d45db04313641d956b7200cd031f23e

src/plugins/lib/ui/index.js
MichalDybizbanskiCreoox: Implemented by https://github.com/xeokit/xeokit-sdk/commit/76590e863203962c526129793de328257d887e81

xeolabs:
@MichalDybizbanskiCreoox do you think it would be doable to make it so that a user can just begin editing any measurement, at any time, by just click-dragging an endpoint? Ie. without explicitly activating the editing control? 

In other words, where clicking/touching an end point would begin editing that measurement.

The reason I ask is because in xeokit BIMViewer (https://github.com/xeokit/xeokit-bim-viewer) the extra "Editing measurement" mode becomes complex UX alongside the BIMViewer's Angle and Distance Measurements tools (which are either active or inactive at any given time) - I can't figure out how to represent the sub-mode for those, for being in "Editing" mode. 

It would be very nice if the user could just drag-drop on any measurement endpoint at any time.

Does it seem like a doable UX to you?  I can imagine though that it would need to differentiate from long-tap, which is used to open context menus.

PR #1576, opened 2024-07-09
XEOK-62 Fix Zone sides winding

MichalDybizbanskiCreoox:
Fixes Zone sides winding, for a correct volume calculation


xeolabs (changes requested):
One small change request - could you please put the plugin in a file called ZonesPlugin.js, and then export that via the index.js file. That's just for consistent code style, where we have each major class in a file that has the same name.

MichalDybizbanskiCreoox:
> One small change request - could you please put the plugin in a file called ZonesPlugin.js, and then export that via the index.js file. That's just for consistent code style, where we have each major class in a file that has the same name.

`ZonesPlugin.js` added.

PR #1574, opened 2024-07-08
Mesh volume and area

xeolabs:

PR #1577, opened 2024-07-10
XEOK-54 Expose SceneModelEntity's area and volume

MichalDybizbanskiCreoox:
This PR introduces `SceneModelEntity::area` and `::volume` getters.
The metrics are computed during load time, and don't rely on run-time geometry information, thus assuming `SceneModelEntity` geometries don't change.
IMPORTANT: The metrics have been implemented for `VBOBatchingTrianglesLayer` and `VBOInstancingTrianglesLayer`, but not for `DTXTrianglesLayer`.

Another PR has been merged to the master branch, which uses a different approach, but provides the same metrics, at
https://github.com/xeokit/xeokit-sdk/pull/1574
so this one becomes a Draft PR for the time being.

### DIFF
--- a/examples/measurement/areaVolume.html
+++ b/examples/measurement/areaVolume.html
@@ -1,0 +1,222 @@
+<!doctype html>
+<html>
+<head>
+    <meta charset="utf-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1">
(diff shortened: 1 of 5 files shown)


tmarti:
Shouldn't volume calculations be only available for entities made exclusively of watertight meshes?

PR #1579, opened 2024-07-11
[FIX] marker occlusion

Kurtil:
There is two parts here:

- If I am correct, `marker.worldPos` = `marker.origin` + marker.rtcPos`. Because the occlusion layers uses the `marker.origin` in the shader, the `marker.rtcPos` should be used to build positions instead of `marker.worldPos`. 

- The commit [XEOK-33](https://github.com/xeokit/xeokit-sdk/commit/1c06e71bf3ae6468cac846cf5fda9816fa7f16e8) seems to introduce a bug in the occlusion tester because "OcclusionLayer markers' transformation is being applied through the OcclusionLayer::positions array" is not correct... ??

### DIFF
--- a/src/viewer/scene/webgl/occlusion/OcclusionLayer.js
+++ b/src/viewer/scene/webgl/occlusion/OcclusionLayer.js
@@ -99,7 +99,7 @@ class OcclusionLayer {
         for (let i = 0; i < this.numMarkers; i++) {
             if (this.markerList[i]) {
                 const marker = this.markerList[i];
-                const worldPos = marker.worldPos;
+                const worldPos = marker.rtcPos;
                 this.positions[j++] = worldPos[0];
                 this.positions[j++] = worldPos[1];
                 this.positions[j++] = worldPos[2];
--- a/src/viewer/scene/webgl/occlusion/OcclusionTester.js
+++ b/src/viewer/scene/webgl/occlusion/OcclusionTester.js
@@ -333,10 +333,7 @@ class OcclusionTester {
                 continue;
             }
 
-            // The `origin` has been changed from `occlusionLayer.origin` to `[ 0, 0, 0 ]`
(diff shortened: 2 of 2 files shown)


Kurtil:
[annotations_createWithMouse.html.zip](https://github.com/user-attachments/files/16176705/annotations_createWithMouse.html.zip)

I've updated the annotations_createWithMouse to quickly test this PR.

Comment/uncomment the lines 70 and/or 93 to verify that it is working correctly :)

paireks:
@MichalDybizbanskiCreoox could you please have a look there? :)

MichalDybizbanskiCreoox:
Thank you @Kurtil for the PR.

The https://github.com/xeokit/xeokit-sdk/commit/1c06e71bf3ae6468cac846cf5fda9816fa7f16e8 commit rectifies an incorrect assignment in the `OcclusionTester` at https://github.com/xeokit/xeokit-sdk/commit/1c06e71bf3ae6468cac846cf5fda9816fa7f16e8#diff-b81e22131b6074c4e31ef6fd6e3cd96cb491eb46ea40497e3beb60b482e400d4
This fix was necessary for a scenario, in which an `Annotation` (which derives from a `Marker`) is created, but doesn't appear in the view if its `worldPos` is set immediately, like this:
`annotations.createAnnotation({ occludable: true }).worldPos = [2,0,-1]`

The example you provide would work before the fix because of the incorrectness being compensated by the original `const worldPos = marker.worldPos;` assignment in the `OcclusionLayer.js` at https://github.com/xeokit/xeokit-sdk/blob/08fe2858b245d8e38d65f4021ff2fb60cec86c21/src/viewer/scene/webgl/occlusion/OcclusionLayer.js#L102

Your change of this original assignment to use `marker.rtcPos` seem to not reintroduce the older error (which is great), and at the same time add precision to the occlusion test being performed with the `rtcPos` property, covering your test case.

PR #1604, opened 2024-07-31
Feature/hatching

ijlal99:
- Users can now instantiate SectionCapsPlugin along with their section planes to enable hatching on the scene
- This plugin supports multiple section planes and models on the scene


src/plugins/SectionCapsPlugin/SectionCapsPlugin.js
```diff
@@ -0,0 +1,362 @@
...
+        this.sceneModel = Object.keys(viewer.scene.models).map((key) => viewer.scene.models[key]);
+        this.prevIntersectionModels = {};
+        this.posTimeout = null;
+        this.dirTimeout = null;
+        this.setupSectionPlanes();
+    }
+
+
+    //this function will be used to setup event listeners for creation of section planes
+    setupSectionPlanes(){
```
xeolabs, line 22: Could we please make any methods private that are only used by SectionCapsPlugin

src/plugins/SectionCapsPlugin/SectionCapsPlugin.js
```diff
@@ -0,0 +1,362 @@
...
+            clearTimeout(this.dirTimeout);
+            this.dirTimeout = null;
+        }, 1000);
+    }
+
+    addHatches(sceneModel, plane) {
+
+        this.deletePreviousModels();
+
+        // const objects = {};
```
xeolabs, line 73: Best remove these commented lines

src/plugins/SectionCapsPlugin/SectionCapsPlugin.js
```diff
@@ -0,0 +1,362 @@
...
+            const planeTransformedVertices = this.transformVertices(vertices, planeTransformMatrix);
+
+            const csgPlane = this.createGeometry(planeTransformedVertices, indices);
+
+            if (csgPlaneMerged)
+                csgPlaneMerged = csgPlaneMerged.union(csgPlane);
+            else
+                csgPlaneMerged = csgPlane;
+        })
+        console.log("csgPlaneMerged: ", csgPlaneMerged);
```
xeolabs, line 203: Let's remove this logging

src/plugins/SectionCapsPlugin/SectionCapsPlugin.js
```diff
@@ -0,0 +1,362 @@
...
+
+        return cappedGeometries;
+    }
+
+    addIntersectedGeometries(csgGometries) {
+        for (const key in csgGometries) {
+            const webglGeometry = this.csgToWebGLGeometry(csgGometries[key]);
+            const model = this.addGeometryToScene(webglGeometry.vertices, webglGeometry.indices, key);
+            if (model)
+                this.prevIntersectionModels[key] = model;
```
xeolabs, line 226: Let's put an underscore in front names of all methods and variables that are only used internally by the plugin, and ensure they're not in the TypeScript definition

src/plugins/SectionCapsPlugin/SectionCapsPlugin.js
xeolabs: Actually, it looks as if almost all methods and variables on SectionCapsPlugin can be made private, since it effectively works in the background. 

I would add something like a `setEnabled(enabled)` method to SectionCapsPlugin, though, so that the user can instantiate the plugin, but still make it inactive temporarily.

xeolabs (changes requested):
Thanks Muhammad! Only changes I'd like to make is to make all methods and variables that are used internally by the plugin private, and remove them from the TS definitions. Then it would seem to have a nice simple API, as an automatic background helper. Also, please add a method, like `setEnabled()`, so the user can disable/enable the plugin dynamically.

src/viewer/Viewer.js
```diff
@@ -3,7 +3,7 @@ import {CameraFlightAnimation} from "./scene/camera/CameraFlightAnimation.js";
 import {CameraControl} from "./scene/CameraControl/CameraControl.js";
 import {MetaScene} from "./metadata/MetaScene.js";
 import {LocaleService} from "./localization/LocaleService.js";
-import html2canvas from 'html2canvas/dist/html2canvas.esm.js';
+import html2canvas from '../../node_modules/html2canvas/dist/html2canvas.esm.js';
```
paireks, line 6: Please remember to remove it at the end, cause later @Kurtil will need to switch it back ;)

src/viewer/scene/model/vbo/instancing/triangles/VBOInstancingTrianglesLayer.js
```diff
@@ -378,7 +378,7 @@ export class VBOInstancingTrianglesLayer {
             && !!textureSet
             && !!textureSet.colorTexture;
 
-        this._state.geometry = null;
+        // this._state.geometry = null;
```
MichalDybizbanskiCreoox, line 381: This removal seems potentially very impactful if there are places that depend on the property test, or by not freeing memory that would otherwise get released.
What would be a reason for the geometry to be kept as the `_state` property, even though the layer was being finalized?

xeolabs: Agreed - better put this line back @ijlal99

ijlal99: I have moved the indices and vertices to 'this._state' and put this line back again.

src/plugins/SectionCapsPlugin/SectionCapsPlugin.js
ijlal99: Done

src/plugins/SectionCapsPlugin/SectionCapsPlugin.js
ijlal99: Done

src/plugins/SectionCapsPlugin/SectionCapsPlugin.js
ijlal99: Done

xeolabs:
Got a strange little artifact when testing examples/slicing/SectionPlanesPlugin_Duplex_SectionCaps.html

Screenshot from 2024-08-07 17-08-22
Screenshot from 2024-08-07 17-08-03

ijlal99:
@xeolabs this is something that is caused by objects that are not solids but classified as solids (I believe).

xeolabs:
> @xeolabs this is something that is caused by objects that are not solids but classified as solids (I believe).

Gotcha - can you make it not make solid caps for non-solid objects?

ijlal99:
@xeolabs that's how I am currently doing it. I am only generating caps for objects that are classified as solids. 

`const isSolid = sceneModel.objects[key].meshes[0].layer.solid ?? false;`

xeolabs:
> @xeolabs that's how I am currently doing it. I am only generating caps for objects that are classified as solids.
> 
> `const isSolid = sceneModel.objects[key].meshes[0].layer.solid ?? false;`

Gotcha - can you avoid this artifact though?

xeolabs:
> @xeolabs that's how I am currently doing it. I am only generating caps for objects that are classified as solids.
> 
> `const isSolid = sceneModel.objects[key].meshes[0].layer.solid ?? false;`

Aha I see. These objects are wrongly classified.

ijlal99:
I am not sure if I can do much about it but would love to hear your ideas.

xeolabs:
Here's an XKT with properly classified solids:

[model.zip](https://github.com/user-attachments/files/16530973/model.zip)


However, it would also be good to have an example that demonstrated solid caps with a programmatically-created SceneModel, like this example: https://xeokit.github.io/xeokit-sdk/examples/scenemodel/#vbo_instancing_autocompressed_triangles

The idea is that we always have a minimal example that's easier to debug (way easier than with an IFC model).

Note that you'd need to change the primitive types on the SceneModel meshes in that example to indicate that they are solids.

ijlal99:
1. I fixed the issue with hatching that was causing some random artifacts to show alongside the hatched model.
2. I have added another parameter with 'opacityThreshold' for user to control hatching the objects that have opacity above this threshold.
3. There was a bug in the implementation of getEachVertex that was not correctly transforming the vertices.
4. I have added a simple example as suggest by @xeolabs but the example with skt model is also there. If you guys want, I can delete it.

Attached is the result of how hatching looks on previous xkt model, the model provided by @xeolabs and a more simpler model.

Screenshot 2024-08-19 at 10 01 26 PM
Screenshot 2024-08-19 at 10 02 19 PM
Screenshot 2024-08-19 at 10 04 01 PM

MichalDybizbanskiCreoox:
I'd suggest reorganizing commits a bit, to avoid introducing unstable state of the repository.
An example would be
https://github.com/xeokit/xeokit-sdk/pull/1604/commits/8acb036bf7bb82105ea7ddf5361bbd9eb8043c07#diff-3986a1ed352df2aeac30a1e1db927c1a2285e4ce219ec61dfa57406f01ef4983
which introduces a `import html2canvas from '../../node_modules/html2canvas/dist/html2canvas.esm.js';` regression, that is rectified by
https://github.com/xeokit/xeokit-sdk/pull/1604/commits/e14358197b9f1df1948e931915f00f6b4992f752
The latter could be merged with the former to prevent introducing a temporary regression to the master branch.

Amoki:
Could it be possible to color or texturize the hatch depending on the IfcType of the object?

PR #1615, opened 2024-08-07
Allow to set different wireframe axes for distance measurements

tmarti:
## Description

This PR allows to externally set the axes of the measurement gizmo when defining a distance measurement.

At the moment, the intented usage is to set it externally:

```js
measurementPlugin.on("measurementCreated", (m) => {
    // e.g. obtain the basis based on the object OBB
    const axesBasis = lookUpObbAlignedBasis(m.origin.entity.id);
    // and use that in the measurement
    m.axesBasis = axesBasis;
});
```

This enables the measurements to be aligned to the measured objects.

## Before

The measurement wires are aligned to the world axes.

image

## After

The measurement wires are aligned to the object OBB.

image

## How it works

Given a concrete basis (by default it's the unit matrix) formed by the columns of the `axesBasis` matrix, in order to calculate the endpoints of the measurement wires (red, green, blue), the code will project the difference-vector in each of the basis' vectors, and use those projections to construct the endpoints.

It's not a breaking change: it will default to same behaviour as today in case the `axesBasis` attribute is not set for the measurement, which means it will stick to world-aligned measurement wires. This is done by setting the unit matrix as default, which will lead to consider the world X-Y-Z axes.

### DIFF
--- a/examples/measurement/distance_createWithMouse_scenegraph_OBB.html
+++ b/examples/measurement/distance_createWithMouse_scenegraph_OBB.html
@@ -1,0 +1,617 @@
+<!doctype html>
+<html>
+<head>
+    <meta charset="utf-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1">
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <title>xeokit Example</title>
+    <link href="../css/pageStyle.css" rel="stylesheet"/>
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/5.13.0/js/all.min.js"></script>
+
+    <style>
+
+        .viewer-ruler-wire-highlighted {
+            border: 2px solid white !important;
+        }
+
+        .viewer-ruler-label-highlighted {
+            border: 2px solid white !important;
+        }
+
+        .viewer-ruler-dot-highlighted {
+            border: 2px solid white !important;
+        }
+
+        .xeokit-context-menu {
+            font-family: 'Roboto', sans-serif;
+            font-size: 15px;
+            display: none;
+            z-index: 300000;
+            background: rgba(255, 255, 255, 0.46);
+            border: 1px solid black;
+            border-radius: 6px;
+            padding: 0;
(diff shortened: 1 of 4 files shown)


xeolabs:
Thanks Toni, could you please make an example for the measurements examples collection (also linked to the example index) that shows how it's done?

So if I understand this correctly, it's aligning the measurements to the OBB of the measured object, not the same as what's sone by the `useRotationAdjustment` configuration(https://github.com/xeokit/xeokit-sdk/blob/master/src/plugins/DistanceMeasurementsPlugin/DistanceMeasurementsPlugin.js#L320)

tmarti:
👌🏻 tomorrow will add the example!

Yes, the purpose of this PR is to align the measurement to the OBB of the object.

For example, if you add a measurement to a wall that is rotated with respect to the world axes, in the case of the OBB-ed measurement, the X axis of the measurement would run along the width of the wall and right over its surface.

(At the moment at least in my case I will be providing the OBB externally, and this PR is an enabler for the e2e mechanism)

I’ve tested the mechanism in a combination of outdoor and indoor cases and works really sweet: the example consuming code above, sets the 1st object selected in the measurement as the one establishing the measurement axis base:

- in the indoor case it works really nice also if you measure between two different objects as usually the objects inside the same room are aligned between them.

- in the outdoor case it also works really well as some models have its perimeter defining a triangle, so this makes sure you’re always measuring across the with/depth of the object OBB.

tmarti:
@xeolabs, done!

I added an example that displays both an unrotated and a rotated scene-graph-based table and that uses the mechanism in the PR description to set the axes for the measurement once it is started.

https://github.com/user-attachments/assets/538904e8-6d3a-4d16-9424-0a9d3fa4b42d

The example is called:

```
examples/measurement/distance_createWithMouse_scenegraph_OBB.html
``` 

Also, made one attribute on `DistanceMeasurement.js` internal (by adding the `_` prefix), and added code-docs to the `axesBasis` getter and setter (both on the TS definitions and the JS class).

xeolabs:
Works well! Thanks very much @tmarti

PR #1623, opened 2024-08-13
XCD-101 Picking overlapping objects

MichalDybizbanskiCreoox:
The PR adds two example pages implementing mouse picking that works with overlapped objects.
One example works by reaching further away with each mouse click.
The other example moves the pick on mouse wheel.


xeolabs:
Merged, thanks @MichalDybizbanskiCreoox  (I'm just killing time before catching a train).
One thing though, which I caught only after merging - could you please move `overlapping.js` into `/examples/libs` and rename it `overlappingPick.js`

MichalDybizbanskiCreoox:
> Merged, thanks @MichalDybizbanskiCreoox (I'm just killing time before catching a train). One thing though, which I caught only after merging - could you please move `overlapping.js` into `/examples/libs` and rename it `overlappingPick.js`

Sure, PR at https://github.com/xeokit/xeokit-sdk/pull/1624

PR #1640, opened 2024-08-23
Add separator as an independent li tag.

ijlal99:
Previously in the context menu, the separator between menu items was achieve by adding inline style of `border-bottom: 1px solid black;`.

Xeovision team needed the separator to be an independent li tag and that makes sense as well. Also, they were struggling with inline css styling that was added to menu items.

I have given the separator an independent li tag.

I have introduced 3 new classes:
1.`xeokit-context-menu-item-seperator` that is assigned to seperators in the menu
2. `xeokit-context-menu-item-visible` for items that are visible.
3. `xeokit-context-menu-item-hidden` for items that are hiddden.

@xeolabs @MichalDybizbanskiCreoox , this change will cause one problem for the users who are currently using context menus that is their hidden menu items will become visible because ContextMenu does not hide or show menu items, instead it toggles between the visible and hidden classes.

They will have to implement the visible and hidden classes like this:
`.xeokit-context-menu-item-visible {
    visibility: visible;
    height: auto;
    padding: null;
}

.xeokit-context-menu-item-hidden {
    visibility: hidden;
    height: 0;
    padding: 0;
}`

Reference ticket: XEOK-92

### DIFF
--- a/examples/buildings/dotbim_BlenderHouse.html
+++ b/examples/buildings/dotbim_BlenderHouse.html
@@ -139,26 +139,25 @@
             padding: 0;
         }
 
-        .xeokit-context-menu ul li {
+        .xeokit-context-menu-item {
             list-style-type: none;
             padding-left: 10px;
             padding-right: 20px;
             padding-top: 4px;
             padding-bottom: 4px;
             color: black;
-            border-bottom: 1px solid gray;
             background: rgba(255, 255, 255, 0.46);
             cursor: pointer;
             width: calc(100% - 30px);
         }
 
-        .xeokit-context-menu ul li:hover {
+        .xeokit-context-menu-item:hover {
             background: black;
             color: white;
             font-weight: normal;
         }
 
-        .xeokit-context-menu ul li span {
+        .xeokit-context-menu-item span {
             display: inline-block;
         }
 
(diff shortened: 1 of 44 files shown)


src/extras/ContextMenu/ContextMenu.js
```diff
@@ -848,14 +854,13 @@ class ContextMenu {
...
-                itemElement.style.height = "0";
-                itemElement.style.padding = "0";
+                itemElement.classList.remove("xeokit-context-menu-item-visible");
+                itemElement.classList.add("xeokit-context-menu-item-hidden");
                 continue;
             } else {
-                itemElement.style.visibility = "visible";
-                itemElement.style.height = "auto";
-                itemElement.style.padding = null;
+                itemElement.classList.remove("ceokit-context-menu-item-hidden");
```
frytaz, line 861: typo

examples/buildings/dotbim_BlenderHouse.html
```diff
@@ -175,6 +174,24 @@
             background: #eeeeee;
             font-weight: normal;
         }
+
+        .xeokit-context-menu-item-seperator {
```
xeolabs, line 178: Typo - rename "seperator" -> "separator"

PR #1657, opened 2024-09-07
Update classes to be documentation friendly

ijlal99:
Updated classes without changing any functionality so that they are documentation friendly.

### DIFF
--- a/src/plugins/DistanceMeasurementsPlugin/DistanceMeasurement.js
+++ b/src/plugins/DistanceMeasurementsPlugin/DistanceMeasurement.js
@@ -240,6 +240,7 @@ class DistanceMeasurement extends Component {
         this._visible = false;
         this._originVisible = false;
         this._targetVisible = false;
+        this._useRotationAdjustment = false;
         this._wireVisible = false;
         this._axisVisible = false;
         this._xAxisVisible = false;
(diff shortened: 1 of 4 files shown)


src/plugins/DistanceMeasurementsPlugin/DistanceMeasurement.js
```diff
@@ -784,6 +785,25 @@ class DistanceMeasurement extends Component {
         return this._targetVisible;
     }
 
+    /**
+     * Sets if the measurement is adjusted based on rotation
+     *
+     * @type {Boolean}
+     */
+    set useRotationAdjustment(value) {
+        value = value !== undefined ? Boolean(value) : this.plugin.useRotationAdjustment;
```
MichalDybizbanskiCreoox, line 794: Using `this.plugin.useRotationAdjustment` as a fallback value of `_useRotationAdjustment` is a pretty significant semantic inconsistency related to other setters - what is the reason to do that?
This setter is a part of an external API, we should be cautious about introducing such unexpected behaviors exposed to the library users.

MichalDybizbanskiCreoox: Ok I see other setters in this particular class are implemented in a similarly peculiar way.

PR #1664, opened 2024-09-10
Feature/hatching

ijlal99:
Hi @MichalDybizbanskiCreoox!

I have implemented somewhat the basics of what we discussed for the hatching branch.

Right now, I am using materials that are already provided by xeokit and I'd suggest that we should keep it that way instead of developing hatching patterns. I am struggling to create different type of materials for each object in the scene model, the baseColor does not seem to be effecting the materials. However, I have confirmed that the sectionCapsPlugin is correctly reading those materials and adding those to section caps.

I would like to suggest that we create a new material type for solids or basic colors.

@xeolabs maybe you can guide me to correctly implement different materials for each object, I am sure I am doing something wrong here.

### DIFF
--- a/examples/slicing/SectionCaps.html
+++ b/examples/slicing/SectionCaps.html
@@ -1,0 +1,196 @@
+<!doctype html>
+<html>
+<head>
+    <meta charset="utf-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1">
+    <meta name="viewport" content="width=device-width, initial-scale=1">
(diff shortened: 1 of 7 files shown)


ijlal99:
@MichalDybizbanskiCreoox please have a look at the PR when you some time now. I believe I have implemented everything apart from calculating intersections with a plane instead of a box.

@xeolabs do you think we can beta release this feature or maybe I should first implement calculation of the intersecting box based on the bounding box of the scene?

PR #1668, opened 2024-09-13
Add submenu position attribute on the li tag containing a submenu

ijlal99:
All those elements in the main context menu, that contain a submenu now store an attribute "data-submenuposition" that tells whether the submenu is position on left or right of the main menu.

### DIFF
--- a/src/extras/ContextMenu/ContextMenu.js
+++ b/src/extras/ContextMenu/ContextMenu.js
@@ -461,6 +461,7 @@ class ContextMenu {
         this._updateItemsTitles();
         this._updateItemsEnabledStatus();
         this._showMenu(this._rootMenu.id, pageX, pageY);
+        this._updateSubMenuInfo();
         this._shown = true;
         this.fire("shown", {});
     }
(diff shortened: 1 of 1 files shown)


src/extras/ContextMenu/ContextMenu.js
```diff
@@ -869,6 +870,22 @@ class ContextMenu {
...
 
+    _updateSubMenuInfo() {
+        if (!this._context) return;
+        let showOnLeft = null;
+        this._itemList.forEach((item) => {
+            if (item.subMenu) {
+                const itemElement = item.itemElement;
+                if (!showOnLeft) {
+                    const itemRect = itemElement.getBoundingClientRect();
+                    const subMenuWidth = 200; // TODO
```
frytaz, line 881: We could set subMenuWidth in ContextMenu cfg ? 
Something like 
`this.subMenuWidth = cfg.subMenuWidth || 200;`
in constructor

ijlal99: I have updated it so that it will automatically read the width of submenu and adjust the position accordingly. No need to provide the width in constructor.

PR #1682, opened 2024-09-27
[FIX] XKTLoaderPlugin returns sceneModel before it is defined

### LINKED ISSUE #1656: XKTLoaderPlugin returns sceneModel before defining it
**Describe the bug**

We noticed that trying to use this library from ClojureScript results in a compilation error when using the [shadow-cljs](https://github.com/thheller/shadow-cljs) compiler, which in turn uses the [Google closure](https://github.com/google/closure-compiler) compiler, which can sometimes be quite strict about the js it consumes. 

```
--- node_modules/@xeokit/xeokit-sdk/dist/xeokit-sdk.cjs.js:133131
Illegal variable reference before declaration: sceneModel
```

I think the root cause is `XKTLoaderPlugin`, where on line [899](https://github.com/xeokit/xeokit-sdk/blob/86ebeb499b730a98d0e21766f26d95fbd00ea464/src/plugins/XKTLoaderPlugin/XKTLoaderPlugin.js#L899) it returns `sceneModel` before defining it (unlike similar code elsewhere). I believe the issue was introduced in [this](https://github.com/xeokit/xeokit-sdk/commit/b21b9b3e42ed16be93df7f751cd2828e3ee64fe2#diff-263b294427d039453337d1b5a9ae202c88a5bda7d143c0fcc6042d86fa1a10f4) commit.

ijlal99:
Fixes #1656

### DIFF
--- a/src/plugins/XKTLoaderPlugin/XKTLoaderPlugin.js
+++ b/src/plugins/XKTLoaderPlugin/XKTLoaderPlugin.js
@@ -894,10 +894,8 @@ class XKTLoaderPlugin extends Plugin {
             delete params.id;
         }
 
-        if (!params.src && !params.xkt && !params.manifestSrc && !params.manifest) {
-            this.error("load() param expected: src, xkt, manifestSrc or manifestData");
-            return sceneModel; // Return new empty model
-        }
+        if (!params.src && !params.xkt && !params.manifestSrc && !params.manifest)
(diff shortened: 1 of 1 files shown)


MichalDybizbanskiCreoox:
I'd consider throwing an error instead of returning an empty scene in this case.
When `XKTLoaderPlugin::load` is used it's rather fair to assume the intention was to load a scene from actual data.
Do you think Invoking this method without any data descriptor could be considered a fatal error instead of a valid call?

PR #1684, opened 2024-09-27
[FEATURE] LASLoaderPlugin point transform options

xeolabs:
Center LAS points as they are loaded:

````javascript
const lasLoaderPlugin = new LASLoaderPlugin(viewer, {
    center: true // Default is false 
});
````

Bake X, Y and Z-axis rotation into LAS points as they are loaded:

````javascript
const lasLoaderPlugin = new LASLoaderPlugin(viewer, {
    rotate: [-90, 0, 0] 
});
````

Bake 90 degree X rotation into LAS points as they are loaded:

````javascript
const lasLoaderPlugin = new LASLoaderPlugin(viewer, {
    rotateX: true 
});
````

Bake transform into LAS points as they are loaded:

````javascript
const lasLoaderPlugin = new LASLoaderPlugin(viewer, {
    transform: [
       1, 0, 0, 0, 
       0, 1, 0, 0, 
       0, 0, 1, 0, 
       0, 0, 0, 1 
    ]
});
````

Then go ahead and load some point clouds:

````javascript
const lasSceneModel = lasLoaderPlugin.load({
    id: "myModel",
    src: "example-no-1.laz"
});
````

Any centering, rotation or transform we've configured on our LASLoaderPlugin gets applied to the point coordinates of all LAS models it now loads.  
  
The plugin applies centering first, then `rotate`, then `rotateX` then `transform`.

In practice, users are not likely to use the combination of `transform` and either `rotate` or `rotateX`, since `transform` could just contain a rotation matrix. The `rotate` and `rotateX` options are shortcuts for common transform cases. 

The SceneModel itself is not transformed as a result of our LasLoaderPlugin configurations (it has zero rotation and position properties), because the plugin transformed the values of the point coordinates before it loaded them into the SceneModel.

### DIFF
--- a/examples/lidar/las_Example1_alignPointCloud.html
+++ b/examples/lidar/las_Example1_alignPointCloud.html
@@ -1,0 +1,238 @@
+<!doctype html>
+<html>
+<head>
+    <meta charset="utf-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1">
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <title>xeokit Example</title>
+    <link href="../css/pageStyle.css" rel="stylesheet"/>
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/5.13.0/js/all.min.js"></script>
+    <script src="../libs/dat.gui.min.js" type="text/javascript"></script>
+    <link href="../css/dat-gui-light-style.css" rel="stylesheet"/>
+    <style>
+
+        /* ----------------------------------------------------------------------------------------------------------*/
+        /* NavCubePlugin */
+        /* ----------------------------------------------------------------------------------------------------------*/
+
+        #myNavCubeCanvas {
+            position: absolute;
+            width: 250px;
+            height: 250px;
+            bottom: 50px;
+            right: 10px;
+            z-index: 200000;
+        }
+
+        .xeokit-camera-pivot-marker {
+            color: #ffffff;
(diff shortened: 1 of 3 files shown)


MichalDybizbanskiCreoox:
Alternatively the API could provide easy ways to define such transformations, to still be passed under a single property.
As in e.g.
`const lasLoaderPlugin = new LASLoaderPlugin(viewer, { matrix: rotate([-90, 0, 0]) });`
or
`const lasLoaderPlugin = new LASLoaderPlugin(viewer, { matrix: rotateX() });`
This doesn't expand the API surface, doesn't introduce confusing semantics about which property is being applied first or even overruled, and on top of that the abstractions can be combined freely by users, or even reused in other contexts.

MichalDybizbanskiCreoox:
What BW compatibility is being considered here?
Isn't the "previous" `rotateX` parameter a member of the `load` method's argument, and not the whole `LASLoaderPlugin`, and also meant to different thing than the "new" `rotateX`?
After this change there's an "old" set of transformation controls that represents model's "runtime" transformation that are specified by the `load` method arguments, and another - a bit similar but different - "new" set of parameters meant to apply a "loadtime" transformation, that is specified by a Plugin's constructor argument.
Is there a reason the "loadtime" transformation is defined for the whole `LASLoaderPlugin`, and not for specific models loads, like the "runtime" transformation is?
Going through the code is looks like there's ten parameters spread across two methods, which all logically represent two operations - a "loadtime" and a "runtime" transformation.

PR #1719, opened 2024-10-31
XCD-144 Fix shaking measurement

MichalDybizbanskiCreoox:
Not sure what was the original intention to hide measurement controls when camera was too close to them, but it causes glitches as described at https://creoox.atlassian.net/browse/XCD-175?focusedCommentId=13492
Removing this special logic seems to fix the reported issue.

### DIFF
--- a/src/plugins/AngleMeasurementsPlugin/AngleMeasurement.js
+++ b/src/plugins/AngleMeasurementsPlugin/AngleMeasurement.js
@@ -1,11 +1,9 @@
-import {Dot3D} from "../lib/ui/index.js";
-import {Wire} from "../lib/html/Wire.js";
-import {Label} from "../lib/html/Label.js";
+import {Dot3D, Label3D, Wire3D} from "../lib/ui/index.js";
 import {math} from "../../viewer/scene/math/math.js";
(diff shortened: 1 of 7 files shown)


xeolabs:
With this PR, what is the behaviour of measurent controls when they intersect the front clipping plane?

MichalDybizbanskiCreoox:
The lines are visible even when they are "behind" the camera, which is no good I guess.
The opposite is not good either though, because if one end of a line is behind the camera, but the other is in front of it, a part of the line should be visible, but that's tricky to achieve with DOM elements.

xeolabs (approved):

ijlal99 (approved):

MichalDybizbanskiCreoox:
The PR introduces a significant change to the architecture of DOM elements (Dot, Wire, Label) visuals management.
Instead of being managed from a plugin abstraction that uses them (e.g. a `DistanceMeasurement`) the elements are now responsible for responding to view-related changes (camera, section planes etc) by themselves.
image

MichalDybizbanskiCreoox:
I've tested the branch with all examples under https://xeokit.github.io/xeokit-sdk/examples/measurement/, all worked correctly.

PR #1728, opened 2024-11-06
Add support for model ids when there are multiple models on the scene

### LINKED ISSUE #1724: [FEATURE] StoreyViewsPlugin option to specify model along with storey ID in various methods

ijlal99:
Closes #1724

### DIFF
--- a/src/plugins/StoreyViewsPlugin/StoreyViewsPlugin.js
+++ b/src/plugins/StoreyViewsPlugin/StoreyViewsPlugin.js
@@ -680,9 +680,10 @@ class StoreyViewsPlugin extends Plugin {
      * @param {Number[]} worldPos 3D World-space position.
      * @returns {String} ID of the storey containing the position, or null if the position falls outside all the storeys.
      */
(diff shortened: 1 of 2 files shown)


src/plugins/StoreyViewsPlugin/StoreyViewsPlugin.js
```diff
@@ -696,9 +697,10 @@ class StoreyViewsPlugin extends Plugin {
      * @param {Number[]} worldPos 3D World-space position.
      * @returns {String} ID of the storey containing the position, or null if the position falls outside all the storeys.
      */
-    getStoreyInVerticalRange(worldPos) {
-        for (let storeyId in this.storeys) {
-            const storey = this.storeys[storeyId];
+    getStoreyInVerticalRange(worldPos, modelId = "") {
```
xeolabs, line 700: Let's make modelId null by default. 
(Even though JS so weirdly equates "" to false!)

xeolabs (changes requested):

PR #1733, opened 2024-11-12
Add ability to capture whole models::StoreyViewsPlugin

### LINKED ISSUE #1725: [FIX] StoreyViewsPlugin should not show sliced models, only whole models

ijlal99:
Fixes #1725


src/plugins/StoreyViewsPlugin/StoreyViewsPlugin.js
```diff
@@ -504,6 +504,7 @@ class StoreyViewsPlugin extends Plugin {
      * @param {Number} [options.width=300] Image width in pixels. Height will be automatically determined from this, if not given.
      * @param {Number} [options.height=300] Image height in pixels, as an alternative to width. Width will be automatically determined from this, if not given.
      * @param {String} [options.format="png"] Image format. Accepted values are "png" and "jpeg".
+     * @param {Boolean} [options.captureWholeStorey=true] Whether the storey map is sliced or not.
```
xeolabs, line 507: Let's rename this to  "captureSectionPlanes", which would have a default value of false, in which case it would be capturing the entire model

PR #1741, opened 2024-11-20
Use elevation property to clip overlapping bounding boxes

ijlal99:


src/plugins/StoreyViewsPlugin/StoreyViewsPlugin.js
```diff
@@ -308,6 +308,31 @@ class StoreyViewsPlugin extends Plugin {
...
+
+            if(isNaN(elevationCur)) return;
+
+            const bb = storeysList[i].storeyAABB;
+            bb[1] = Math.max(bb[1], parseFloat(elevationCur) / 1000);
+
+            if (i > 0) {
+                const storeyMetaObjectNext = metaScene.metaObjects[storeysList[i - 1].storeyId];
+                const elevationNext = storeyMetaObjectNext.attributes.elevation;
+                bb[4] = Math.min(bb[4], parseFloat(elevationNext) / 1000);
```
xeolabs, line 330: Let's make this work for all "up" directions - you could use Camera worldUp, xUp, yUp, zUp etc to select which element of bb to update

PR #1765, opened 2024-12-20
Implement missing cross section from dtx renderer

ijlal99:
@xeolabs can you please have a look just in case I missed any dtx renderer.


paireks (commented):
LGTM, have you tried it with this example (but modified to use DTX if possible?): https://xeokit.github.io/xeokit-sdk/examples/slicing/#SectionPlanesPlugin_Duplex_SectionPath_CrossSections

ijlal99:
@paireks thank you for your input. I did check the mentioned example with dtx enabled and it works.

PR #1775, opened 2025-01-14
Implement section caps through sceneModelEntity:capMaterial

ijlal99:
Please give this branch a test and let me know if you see any bugs so that we can fix it before merging into main.

### DIFF
--- a/examples/slicing/SectionPlanesPlugin_Duplex_SectionCaps.html
+++ b/examples/slicing/SectionPlanesPlugin_Duplex_SectionCaps.html
@@ -1,0 +1,159 @@
+<!doctype html>
+<html>
+
+<head>
+    <meta charset="utf-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1">
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <title>xeokit Example</title>
+    <link href="../css/pageStyle.css" rel="stylesheet" />
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/5.13.0/js/all.min.js"></script>
+</head>
+
+<body>
+    <input type="checkbox" id="info-button" />
+    <label for="info-button" class="info-button"><i class="far fa-3x fa-question-circle"></i></label>
+    <canvas id="myCanvas"></canvas>
+    <div class="slideout-sidebar">
+        <img class="info-icon" src="../../assets/images/section_plane_icon.png" />
+        <h1>SectionPlanesPlugin</h1>
+        <h2>Slices models open to reveal internal structures</h2>
+        <p>In this example, we're loading an IFC2x3 BIM model from the file system, then slicing it with a section
+            plane.</p>
+        <h3>Stats</h3>
+        <ul>
+            <li>
+                <div id="time">Loading JavaScript modules...</div>
+            </li>
+        </ul>
+        <h3>Components used</h3>
+        <ul>
+            <li>
+                <a href="../../docs/class/src/viewer/Viewer.js~Viewer.html" target="_other">Viewer</a>
+            </li>
+            <li>
+                <a href="../../docs/class/src/plugins/XKTLoaderPlugin/XKTLoaderPlugin.js~XKTLoaderPlugin.html"
+                    target="_other">XKTLoaderPlugin</a>
+            </li>
+            <li>
+                <a href="../../docs/class/src/plugins/SectionPlanesPlugin/SectionPlanesPlugin.js~SectionPlanesPlugin.html"
+                    target="_other">SectionPlanesPlugin</a>
+            </li>
+        </ul>
+        <h3>Resources</h3>
+        <ul>
+            <li>
+                <a href="http://openifcmodel.cs.auckland.ac.nz/Model/Details/274" target="_other">Model source</a>
+            </li>
+        </ul>
+    </div>
+</body>
+
+<script type="module">
+
+    //------------------------------------------------------------------------------------------------------------------
+    // Import the modules we need for this example
+    //------------------------------------------------------------------------------------------------------------------
+
+    import { Viewer, SceneModel, XKTLoaderPlugin, SectionPlanesPlugin, PhongMaterial, math } from "../../dist/xeokit-sdk.min.es.js";
+
+    //------------------------------------------------------------------------------------------------------------------
+    // Create a Viewer and arrange the camera
+    //------------------------------------------------------------------------------------------------------------------
+
+    const viewer = new Viewer({
+        canvasId: "myCanvas",
+        transparent: true,
+        readableGeometryEnabled: true
+    });
+
+    viewer.camera.eye = [-2.341298674548419, 22.43987089731119, 7.236688436028655];
+    viewer.camera.look = [4.399999999999963, 3.7240000000000606, 8.899000000000006];
+    viewer.camera.up = [0.9102954845584759, 0.34781746407929504, 0.22446635042673466];
+
+    const cameraControl = viewer.cameraControl;
+    cameraControl.navMode = "orbit";
+    cameraControl.followPointer = true;
+
+    //----------------------------------------------------------------------------------------------------------------------
+    // Create a xeokit loader plugin, load a model, fit to view
+    //----------------------------------------------------------------------------------------------------------------------
+
+    const xktLoader = new XKTLoaderPlugin(viewer);
+
+    var t0 = performance.now();
+
+    document.getElementById("time").innerHTML = "Loading model...";
+
+    const sceneModel = xktLoader.load({
+        id: "myModel",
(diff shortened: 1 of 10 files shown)


paireks:
Ok, I will give it a try :)

paireks (commented):
Hey guys, I'm testing it on various different models, it works better than I thought. It runs quite fast and gets quite good results. But also it has some limitations:
- models have to have all meshes closed (solids). But in reality many models have none of them closed or only some of them closed, which results in only partial caps / no caps.
- for huge models (such as WestRiverSideHospital) it crashes my browser.
- it doesn't work once I set the position of the model (it doesn't cap then)
- I think it would be good to check the model far from origin, but I can't remember which one was it. Do you remember @xeolabs?

Because of the above I'm thinking if maybe it would be better to add this functionality more as an example rather than a xeokit-sdk functionality? What do you guys think?

ijlal99:
Hi @paireks!

Thank you for trying out this feature and your valuable feedback. 

1. For the capping of only solid objects, this is not really a limitation but actually a requirement that we only hatch solid objects. https://github.com/xeokit/xeokit-sdk/blob/d324265afae2c32adc10601c6da0132ee40b6129/src/viewer/scene/sectionCaps/SectionCaps.js#L201
2. For huge models like WestRiverSideHospital, I checked and the browser crashes even when the SectionCaps are disabled.
3. We currently do not have any event in xeokit that informs us that the position of an object has been changed. We plan to handle this situation later.
4. Our first implementation of this feature was a plugin and then we switched and made it a part of xeokit sdk with the agreement of @MichalDybizbanskiCreoox and @xeolabs.

MichalDybizbanskiCreoox:
> 2\. For huge models like WestRiverSideHospital, I checked and the browser crashes even when the SectionCaps are disabled.

I've added a section plane and capMaterials to the `xkt_dtx_WestRiverSideHospital.html` example (also calling `Viewer` constructor with the `readableGeometryEnabled: true` flagg needed for hatching).
The example works for me in Chrome on macOS just fine: 
https://github.com/user-attachments/assets/9dfe8b42-c5a3-4c7e-baa5-ecc133d5355b

src/viewer/scene/sectionCaps/SectionCaps.js
```diff
@@ -0,0 +1,810 @@
...
+                    delete this._verticesMap[key];
+                    delete this._indicesMap[key];
+                }
+            }
+            this._update();
+        })
+        
+    }
+
+    _onCapMaterialUpdated(entityId, modelId) {
```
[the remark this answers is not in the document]
ijlal99: I have moved the geometry calculation to _getIntersectionSegments and only calculated it if the object is being intersected. Thank you!

src/viewer/scene/sectionCaps/SectionCaps.js
```diff
@@ -0,0 +1,810 @@
...
+                        mesh.getEachIndex((_indices) => {
+                            indices[index].push(_indices);
+                        })
+                        index++;
+                    }
+
+                })
+            }
+            else {
+                entity.getEachVertex((_vertices) => {
```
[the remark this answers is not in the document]
ijlal99: Thank you for pointing this out. This approach was required previously when we were working with csg, to reduce the number of random artifacts. I totally forgot about removing it.

src/viewer/scene/sectionCaps/SectionCaps.js
```diff
@@ -0,0 +1,800 @@
...
+                    delete this._indicesMap[key];
+                    this._update();
+                }
+            }
+        })
+        
+    }
+
+    _onCapMaterialUpdated(entityId, modelId) {
+        if(!this.scene.readableGeometryEnabled) return;
```
MichalDybizbanskiCreoox, line 76: This test is probably unnecessary, since the requirement is already being verified by the constructor.

MichalDybizbanskiCreoox: I see, that makes sense, thanks!
Could we then consider removing the test from the constructor?
I assume it's there to prevent using the resources for the case of a scene that doesn't enable readable geometries, so cannot use the capMaterials anyway.
There will be use cases however of scenes that do enable readable geometries, but won't use capMaterials anyway, and those cases would be taxed by the resources usage.
Perhaps the initialization that's currently included inside the constructor could be moved to be lazily executed only when the `_onCapMaterialUpdated` is called the first time (given the `this.scene.readableGeometryEnabled` test is satisfied).
This way there's no redundancy in the condition (which could change in the future), and on top of that unless a user doesn't use capMaterials, they won't have to pay the price for the feature.

ijlal99: Thank you for your input, Michal! That totally makes sense. I'll make these changes.

examples/slicing/SectionPlanesPlugin_Duplex_SectionCaps.html
```diff
@@ -0,0 +1,159 @@
...
+
+        //------------------------------------------------------------------------------------------------------------------
+        // Add caps materials to all objects inside the loaded model that have an opacity equal to or above 0.7
+        //------------------------------------------------------------------------------------------------------------------        
+        const opacityThreshold = 0.7;
+        const material = new PhongMaterial(viewer.scene,{
+            diffuse: [1.0, 0.0, 0.0],
+            backfaces: true
+        });
+        addCapsMaterialsToAllObjects(sceneModel, opacityThreshold, material);
```
MichalDybizbanskiCreoox, line 107: When the immediate `addCapsMaterialsToAllObjects(...)` call is removed, and instead called after the model loads (from e.g. a test function `window.addHatching = () => addCapsMaterialsToAllObjects(...);` the hatching doesn't appear.

src/viewer/scene/model/SceneModelEntity.js
```diff
@@ -649,6 +651,30 @@ export class SceneModelEntity {
...
     }
 
+    /**
+     * Sets the SceneModelEntity's capMaterial that will be used on the caps generated when this entity is sliced
+     *
+     * Default value is ````null````.
+     *
+     * @type {Material}
+     */
+    set capMaterial(value) {
```
xeolabs, line 661: I have one architectural change request:

- add a boolean `Entity.capped` property, which would indicate if that `Entity` is currently capped
- add a `Scene.capMaterial` property, which is a `Material`
- then make `Entity.capMaterial` point to `Scene.capMaterial` unless overridden by a user's `Material` instance

This way, we're consistent with selection, highlighting etc. 

By default, the `Material` would be `Scene.capMaterial`, unless it's assigned to a user's custom

src/viewer/scene/sectionCaps/SectionCaps.js
```diff
@@ -0,0 +1,675 @@
...
+
+function pointsEqual(p1, p2) {
+    return (
+        Math.abs(p1[0] - p2[0]) < epsilon &&
+        Math.abs(p1[1] - p2[1]) < epsilon &&
+        Math.abs(p1[2] - p2[2]) < epsilon
+    );
+}
+
+class SectionCaps {
```
xeolabs, line 24: Need a class comment - maybe something similar to [`SAO`](https://github.com/xeokit/xeokit-sdk/blob/master/src/viewer/scene/postfx/SAO.js)

xeolabs (changes requested):

src/viewer/scene/model/SceneModelEntity.js
MichalDybizbanskiCreoox: At the moment semantics of not assigning `capMaterial` are: "don't create section caps for this entity", which is useful for users that are not interested in hatching, and so don't have to pay the performance price for the feature.
It's also as simple as a single line of code to apply any global/default `capMaterial` to all entities.

src/viewer/scene/sectionCaps/SectionCaps.js
```diff
@@ -0,0 +1,675 @@
...
+
+                    if(!this._verticesMap[sceneModel.id]) {
+                        this._verticesMap[sceneModel.id] = new Map();
+                        this._indicesMap[sceneModel.id] = new Map();
+                    }
+
+                    let vertices = [], indices = [];
+
+                    if(!this._verticesMap[sceneModel.id].has(key)) {
+                        const isSolid = object.meshes[0].layer.solid !== false;
```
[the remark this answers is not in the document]
MichalDybizbanskiCreoox: A `primitive` has different semantics throughout the system.
Some places consider it as e.g. `solid` or `surface` (`SceneModelMesh`), while others consider it as e.g. `triangle-strip` or `line-loop` (`ReadableGeometry`).
Perhaps it'd be clearer, if the `SceneModelMesh` exposed a `isSolid()` predicate, instead of a confusing `primitive` property?

MichalDybizbanskiCreoox:
I've checked the branch, including locally rebased to `master`, looks ok to me.

xeolabs:
Thanks @ijlal99 , could you please add a description to this PR, for users checking the change log etc

PR #1788, opened 2025-01-23
Much cleaner edge rendering when enabling logarithmic depth buffer

tmarti:
## Descrption

Use the same gradient technique has here https://github.com/xeokit/xeokit-sdk/issues/1162#issuecomment-1778744218 for much better visual quality in edge rendering when the logarithmic depth buffer is enabled.

## Screenshots: Before / After

The screenshots below were taken with a far plane distance equal to 1M km!

### Before (i)

image

### After (i)

image

### Before (ii)

image

### After (ii)

image

## Explanation

Sub-pixel z-fighting (explained here https://github.com/xeokit/xeokit-sdk/issues/1162#issuecomment-1774049269) makes edge drawing worse when logarithmic depth buffer is enabled, possibly because the dynamic range of the Z values get compressed (via the `log` function) together with much bigger far plane distances being used with logarithmic z-buffer.

As the rendering of edged surfaces happens in two stages in xeokit-sdk...

- a) first, regular surfaces are rendered, and this initializes the Z-buffer
- b) on top of it, edges are rendered, which then z-fight with the pre-initialized z-buffer in "a)"

... the reduced dynamic range affects more the case where log-depth-buffer is enabled.

This PR makes use of the gradient trick so, during the regular surface rendering, their z-values are "pushed" far away from the camera to the same distance where their immediate furthest neighbour pixel is.

This makes the z-buffer pre-initialization much more friendly with stage "b)", and fixes the sub-pixel z-fighting problem.

### DIFF
--- a/src/viewer/scene/model/vbo/batching/triangles/renderers/TrianglesColorRenderer.js
+++ b/src/viewer/scene/model/vbo/batching/triangles/renderers/TrianglesColorRenderer.js
@@ -221,7 +221,10 @@ export class TrianglesColorRenderer extends TrianglesBatchingRenderer {
         }
 
         if (scene.logarithmicDepthBufferEnabled) {
-            src.push("    gl_FragDepth = isPerspective == 0.0 ? gl_FragCoord.z : log2( vFragDepth ) * logDepthBufFC * 0.5;");
+            src.push("    float dx = dFdx(vFragDepth);")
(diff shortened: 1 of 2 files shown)


tmarti:
Just checked, it should indeed fix it 🎁

pbbr:
@tmarti Very nice improvement, congrats! Do you see a way to improve the edge rendering in ortho / 2D mode too?

PR #1806, opened 2025-02-11
Update package.json

drashevski:
updated description & keywords for better SEO on npm


drashevski:
removed "git+" from repo url (causes probably snyk to not see the actual number of contributors)
updated license to show as AGPL-3.0

PR #1829, opened 2025-03-07
Implement mat4ToMat3 in math.js

MathiasWP:
just noticed that it was missing, figured i could implement it.

i followed row‐major


xeolabs:
Thanks @MathiasWP !

PR #1833, opened 2025-03-11
Embed xeokit version number for identification

ijlal99:
@xeolabs please have a look at this PR that:
1. Prints xeokit version number being used in the console
2. Add xeokit version number as a dataset on the canvas being used.

What I did to achieve this:
1. Created a placeholder string in viewer.js file.
2. This string is replaced by the origin version number taken from package.json, when the project is built.
3. We output this string in Viewer.js and also supply this string to Scene.js that sets this string as a dataset value on the canvas being used.

Please let me know if you think this is not the right way or you have a better approach in mind.

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -1,5 +1,9 @@
 import {nodeResolve} from '@rollup/plugin-node-resolve';
 import { getBabelOutputPlugin } from '@rollup/plugin-babel';
+import replace from '@rollup/plugin-replace';
+import { readFileSync } from 'fs';
+
+const pkg = JSON.parse(readFileSync('./package.json', 'utf8'));
 
 export default {
     input: './src/index.js',
@@ -29,6 +33,11 @@ export default {
         nodeResolve({
             browser: true,
             preferBuiltins: false
-        })
+        }),
+        replace({
+            delimiters: ['', ''],
+            '__VIEWER_VERSION__': JSON.stringify(pkg.version),
+            preventAssignment: true
+        }),
     ]
 }
--- a/src/viewer/Viewer.js
+++ b/src/viewer/Viewer.js
@@ -7,6 +7,8 @@ import html2canvas from 'html2canvas/dist/html2canvas.esm.js';
 import {math} from "./scene/math/math.js";
 import {transformToNode} from "../plugins/lib/ui/index.js";
 
+const VIEWER_VERSION = '__VIEWER_VERSION__';
+
 /**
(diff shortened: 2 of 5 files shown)


xeolabs:
Nice, that's an elegant way to do it, thanks @ijlal99

src/viewer/Viewer.js
```diff
@@ -68,6 +70,10 @@ class Viewer {
      */
     constructor(cfg) {
 
+        this.version = VIEWER_VERSION;
+
+        console.info('Xeokit SDK Version: ', this.version);
```
MichalDybizbanskiCreoox, line 75: Wouldn't it be enough to just expose the version number, skipping the console output?

rollup.config.js
```diff
@@ -29,6 +33,11 @@ export default {
         nodeResolve({
             browser: true,
             preferBuiltins: false
-        })
+        }),
+        replace({
+            delimiters: ['', ''],
+            '__VIEWER_VERSION__': JSON.stringify(pkg.version),
```
MichalDybizbanskiCreoox, line 39: Shouldn't the pattern to replace be `'\'__VIEWER_VERSION__\''` instead of `'__VIEWER_VERSION__'`?

src/viewer/Viewer.js
ijlal99: That would be enough but having a console output would make it easy to identify the version number.

rollup.config.js
ijlal99: We want to replace the value inside the string at:
const VIEWER_VERSION = '__VIEWER_VERSION__';

so we make rollup replace plugin to find the exact string i.e. __VIEWER_VERSION__ and replace it with the actual version number. If we add backslashes then replace plugin will for an exact match for: \'__VIEWER_VERSION__\'

What I am thinking from your comment is that you want to use escape characters so that we replace the version string along with it's quotes but replace looks for exact match rather than a pattern. Or maybe you are trying to make a different point here?

MichalDybizbanskiCreoox: Let the `package.json` contain ```... "version": "2.6.70"  ...```.
The contents of a built file will be
```const VIEWER_VERSION = '"2.6.70"';```
but shouldn't they be
```const VIEWER_VERSION = '2.6.70';```
instead?

xeolabs: Ah true - I missed that - there are quotes in the version string itself.

@ijlal99 could you please make it so that the quotes are removed.

Issue #1704, opened 2024-10-21
Large number of annotations lagging in the latest version


hamza-hajji:
**Describe the bug**
In xeokit sdk 2.6.10, rendering 200 annotations runs smoothly, in xeokit-sdk@2.6.50, there is significant lagging, I put two screen casts of the 200 annotations in different versions

**To Reproduce**

Render 200 annotations in xeokit-sdk@2.6.50

```js
const annotations = new AnnotationsPlugin(viewer, {
  // Default HTML template for marker position
  markerHTML:
    "<div class='annotation-marker' style='background-color: {{markerBGColor}};'>{{glyph}}</div>",

  // Default HTML template for label
  labelHTML:
    "<div class='annotation-label' style='background-color: {{labelBGColor}};'>" +
    "<div class='annotation-title'>{{title}}</div><div class='annotation-desc'>{{description}}</div></div>",

  // Default values to insert into the marker and label templates
  values: {
    markerBGColor: "red",
    labelBGColor: "red",
    glyph: "X",
    title: "Untitled",
    description: "No description",
  },
});

for (let index = 0; index < 200; index++) {
  annotations.createAnnotation({
    id: "myAnnotation" + index,
    worldPos: [Math.random() * 2, Math.random() * 2, Math.random() * 2], // 3D World-space position
    occludable: false, // Optional, default, makes Annotation invisible when occluded by Entities
    markerShown: true, // Optional, default is true, makes position visible (when not occluded)
    labelShown: false, // Optional, default is false, makes label visible (when not occluded)
    values: {
      // Optional, overrides AnnotationPlugin's defaults
      glyph: "O",
      title: "Origin",
      description: "This is my annotation.",
      markerBGColor: "blue",
    },
  });
}

```

![200AnnotationsLag](https://github.com/user-attachments/assets/5a05e002-8c4d-4af1-a8ae-951ffa4c1ce4)

It lags

**Expected behavior**
It doesn't lag, same behavior as xeokit-sdk@2.6.10

![ezgif-5-372170b6fa](https://github.com/user-attachments/assets/e2005b8e-7570-4a34-90bc-eff9de61accc)

MichalDybizbanskiCreoox:
Hi @hamza-hajji, thank you for your report!

[...]

PR #1841, opened 2025-03-25
XCD-267 Implement "legacy" annotation alignment for the large number of annotations case

MichalDybizbanskiCreoox:
If an `Annotation::setMarkerAlign` is called with the `legacy` argument, the annotation should position as it used to before the "XEOK-44 Expose Annotation::setMarkerAlign and Annotation::setLabelPosition" commit https://github.com/xeokit/xeokit-sdk/commit/a70545c7c0b569c317c6e00897444d09a5e353f1


MichalDybizbanskiCreoox:
Also related to #1704

PR #1846, opened 2025-04-01
add optional parameter 'rootDOMNode' to allow working with ShadowRoot

annawilczura:
Changes from this PR are related to creating web components using viewer and related html elements, where they are inside of ShadowRoot.
Elements that are nested inside of ShadowRoot element tree cannot be searched by querying document object, because they are isolated from the rest of the DOM. Also globally set CSS classes don't affect them.
Here I propose an optional 'rootDOMNode' parameter, to which we can pass the ShadowRoot and then append children to it or query the root element instead doing it on document object. The parameter can be used in NavCubePlugin (and it's TextureCanvas), Scene and ContextMenu configs.

### DIFF
--- a/src/extras/ContextMenu/ContextMenu.js
+++ b/src/extras/ContextMenu/ContextMenu.js
@@ -282,6 +282,7 @@ class ContextMenu {
      * @param {Boolean} [cfg.enabled=true] Whether this ````ContextMenu```` is initially enabled. {@link ContextMenu#show} does nothing while this is ````false````.
      * @param {Boolean} [cfg.hideOnMouseDown=true] Whether this ````ContextMenu```` automatically hides whenever we mouse-down or tap anywhere in the page.
      * @param {Boolean} [cfg.hideOnAction=true] Whether this ````ContextMenu```` automatically hides after we select a menu item. Se false if we want the menu to remain shown and show any updates to its item titles, after we've selected an item.
+     * @param {Node | undefined} [cfg.rootDOMNode] Optional reference of an existing DOM Node (e.g. ShadowRoot), which encapsulates all HTML elements related to viewer plugins, defaults to ````document.body````.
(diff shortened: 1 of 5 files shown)


src/extras/ContextMenu/ContextMenu.js
```diff
@@ -282,6 +282,7 @@ class ContextMenu {
      * @param {Boolean} [cfg.enabled=true] Whether this ````ContextMenu```` is initially enabled. {@link ContextMenu#show} does nothing while this is ````false````.
      * @param {Boolean} [cfg.hideOnMouseDown=true] Whether this ````ContextMenu```` automatically hides whenever we mouse-down or tap anywhere in the page.
      * @param {Boolean} [cfg.hideOnAction=true] Whether this ````ContextMenu```` automatically hides after we select a menu item. Se false if we want the menu to remain shown and show any updates to its item titles, after we've selected an item.
+     * @param {Node | undefined} [cfg.rootDOMNode] Optional reference of an existing DOM Node (e.g. ShadowRoot), which encapsulates all HTML elements related to viewer plugins, defaults to ````document.body````.
```
MichalDybizbanskiCreoox, line 285: I suggest renaming all `rootDOMNode` to `parentElement` occurrences in this commit, for two reasons:
1. There are already cases of this kind of a constructor's argument (albeit non necessarily named properties, but positional) that are named `parentElement`, so we'd retain some consistency, without introducing new symbols.
2. The parameter appears to have exactly the same semantics and purpose as the DOM's `Node::parentElement` property, so using the same name makes its purpose more explanatory and evident than otherwise.

MichalDybizbanskiCreoox: Depending on specific argument requirements the `parentElement` could be named `parentNode` instead.

PR #1856, opened 2025-04-10
Feature/context menu position relative to offset parent

annawilczura:
Changes will allow positioning of ````ContextMenu```` relative to viewer canvas' container instead of 'window'. Also supports viewer canvases not attached to the top of the document.
Additionally I've changed ````_clear()```` method to remove menuElement directly to prevent errors in some cases.

### DIFF
--- a/src/extras/ContextMenu/ContextMenu.js
+++ b/src/extras/ContextMenu/ContextMenu.js
@@ -753,15 +753,15 @@ class ContextMenu {
                             const showOnLeft = (itemRect.left - subMenuWidth) > offsetRect.left;
                         
                             if(showOnRight)
-                                self._showMenu(subMenu.id, itemRect.right - 5, itemRect.top - 1);
+                                self._showMenu(subMenu.id, itemRect.right - 5, itemRect.top + window.scrollY - 1);
                             else if (showOnLeft)
-                                self._showMenu(subMenu.id, itemRect.left - subMenuWidth, itemRect.top - 1);
+                                self._showMenu(subMenu.id, itemRect.left - subMenuWidth, itemRect.top + window.scrollY - 1);
                             else {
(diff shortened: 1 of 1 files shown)


src/extras/ContextMenu/ContextMenu.js
```diff
@@ -945,14 +954,20 @@ class ContextMenu {
...
+
+        const offsetRect = this._offsetParent.getBoundingClientRect();
+        
+        const bottomContainerBorder = offsetRect.bottom + window.scrollY;
+        if ((pageY + menuHeight) > bottomContainerBorder) {
+            pageY = bottomContainerBorder- menuHeight;
         }
-        if ((pageX + menuWidth) > window.innerWidth) {
-            pageX = window.innerWidth - menuWidth;
+        if ((pageX + menuWidth) > offsetRect.right) {
```
MichalDybizbanskiCreoox, line 964: Haven't tried it, but if `window.scrollY` is considered for `pageY`, wouldn't it be worth it to also consider `window.scrollX` for `pageX`?

annawilczura: Yeah, assuming that horizontal scroll is generally unwanted situation, but I'll add it just in case.

MichalDybizbanskiCreoox: Looks like the assignment should be `pageX = rightContainerBorder - menuWidth;` right?

annawilczura:
I still need to test the scrollX for submenus

annawilczura:
I've tested already on webcomponent, as well as xeokit-bim-viewer app (also embedded with iframe), seems to work fine and I'm not going to add anything more to this PR, so please feel free to merge if you don't find any more issues here :)

PR #1857, opened 2025-04-10
Support section handle responsiveness on viewer canvas placed on non-top position of scrolled document

annawilczura:
SectionPlanesPlugin' handles were unresponsive to mouse events when viewer container was placed in document with some other content above. This change fixes it.


MichalDybizbanskiCreoox:
Good catch.
There's a lot of UI around the project that relies on `clientX` and `clientY`.
I suggest we create a JIRA ticket to investigate and fix if needed all cases, with an example that exposes the problem.

PR #1858, opened 2025-04-11
Hide elements with the specific type during navigation [Prototype]

paireks:
Hello guys,

this is a prototype of showing/hiding elements of specific type during navigation for performance (method suggested by Lindsay).

You can see below it hides the trees in that specific example:
2025-04-11_14h25_39

I think it should be tested on a much heavier model to compare the results.

Any comments and suggestions highly apprieciated!

### DIFF
--- a/examples/performance/FastNavPlugin_HolterTower.html
+++ b/examples/performance/FastNavPlugin_HolterTower.html
@@ -98,7 +98,8 @@
         defaultScaleCanvasResolutionFactor: 1.0,
         scaleCanvasResolutionFactor: 0.5,
         delayBeforeRestore: true,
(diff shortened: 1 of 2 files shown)


paireks:
It looks like there is some speed boost looking at this model, before:

https://github.com/user-attachments/assets/34572f9d-5fbc-4880-9719-0710636c91d8

After:

https://github.com/user-attachments/assets/2b1031f2-7b94-40e3-8175-d629f1a1d468

paireks:
Additional example where it might especially makes sense, like models with reinforcement

https://github.com/user-attachments/assets/e2688a28-284e-4e5d-85d3-8f7a325992f1

PR #1861, opened 2025-04-14
[FIX] Use color and opacity for LineSet

paireks:
This PR introduces 3 separate changes:
1. It uses color for LineSet:
image
2. It changes the default color to [1,1,1] (white), to not change the current behaviour (right now it was all white no matter if there was a color).
3. It uses also opacity for LineSet:
2025-04-14_14h55_56

Looking for the feedback :)


paireks:
Thank you for help and check! Should be ready to be merged.

PR #1863, opened 2025-04-15
[FEATURE] Add callback functions to FastNavPlugin.js with 3 different examples

paireks:
Continuation of https://github.com/xeokit/xeokit-sdk/pull/1858

It looks to me that adding these callbacks as @MichalDybizbanskiCreoox suggested is much better than adding our own logic in the xeokit. Please check if this is what you expected.

This optimization feature has 3 additional examples:
1. With ignoring elements of specific type:
2025-04-15_16h16_17

2. With showing only elements of specific type:
2025-04-15_16h17_03

3. With ignoring elements which have lots of triangles:
2025-04-15_16h17_49

It can be much more general this way and gives more power to the user.

Please review before merging. Types will be updated after this part is accepted.

### DIFF
--- a/examples/performance/FastNavPlugin_hideElementsWithComplexGeometry.html
+++ b/examples/performance/FastNavPlugin_hideElementsWithComplexGeometry.html
@@ -1,0 +1,294 @@
+<!doctype html>
+<html>
+<head>
+    <meta charset="utf-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1">
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <title>xeokit Example</title>
+    <link href="../css/pageStyle.css" rel="stylesheet"/>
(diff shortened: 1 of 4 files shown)


examples/performance/FastNavPlugin_hideElementsWithComplexGeometry.html
```diff
@@ -0,0 +1,296 @@
...
+        hideSAO: true,
+        hideColorTexture: true,
+        hidePBR: true,
+        hideTransparentObjects: false,
+        scaleCanvasResolution: true,
+        defaultScaleCanvasResolutionFactor: 1.0,
+        scaleCanvasResolutionFactor: 0.5,
+        delayBeforeRestore: true,
+        delayBeforeRestoreSeconds: 0.4,
+        callbackForLowQuality: hideElementsWithComplexGeometry,
```
xeolabs, line 131: I would rename 

callbackForLowQuality -> onMoving
callbackForHighQuality -> onStopped

PR #1871, opened 2025-04-27
XEOK-107: Add ability to render 360 images using spheremaps

ijlal99:


src/plugins/SphereMapsPlugin/SphereMapsPlugin.js
MichalDybizbanskiCreoox, line 1: What's the point of the `SphereMapsPlugin`?
It looks like just a hashmap for `SphereMap`s.

src/viewer/scene/spheremap/SphereMap.js
MichalDybizbanskiCreoox, line 1: Do we need a class for a `SphereMap`?
After all it's only a `Mesh` including a textured material and a geometry.
Could we consider providing instead a "helper" function outside of the `viewer` hierarchy, that returns a "SphereMap" `Mesh`, which users can then manage themselves like any other mesh, without introducing an artificial level of abstraction?
There's a lot of details hardcoded in the class, which might give an impression they have to be like that, while this implementation is one of many ways such a feature can be implemented (for example consider a SphereMap made of already allocated textures).

MichalDybizbanskiCreoox (changes requested):

src/viewer/scene/spheremap/SphereMap.js
xeolabs: I agree with this recommendation - a non-core factory function that creates Node and Mesh components, no special classes.

examples/viewer/sphere_map.html
```diff
@@ -0,0 +1,126 @@
...
+        </ul>
+    </div>
+</body>
+<script type="module">
+
+    //------------------------------------------------------------------------------------------------------------------
+    // Import the modules we need for this example
+    //------------------------------------------------------------------------------------------------------------------
+
+    import { Viewer, XKTLoaderPlugin, SphereMapsPlugin } from "../../dist/xeokit-sdk.min.es.js";
```
paireks, line 81: XKTLoaderPlugin can be removed

examples/viewer/sphere_map.html
```diff
@@ -0,0 +1,126 @@
...
+            <li>
+                <a href="../../docs/class/src/plugins/SphereMapsPlugin/SphereMapsPlugin.js~SphereMapsPlugin.html"
+                    target="_other">SphereMapsPlugin</a>
+            </li>
+        </ul>
+
+        <h3>Assets</h3>
+        <ul>
+            <li>
+                <a href="https://www.flickr.com/photos/jonragnarsson/2294472375/" target="_other">Abandoned Equirectangular by Jón Ragnarsson</a>
```
paireks, line 67: Please double check the licenses of both of them to be 100% sure if we can use them and if this kind of reference is enough for them both.

paireks (changes requested):
One additional question related to the behaviour of this example:

Shouldn't it move in the opposite direction? That was my instinct once I was dragging the mouse to move it around. But maybe it's just me.

examples/viewer/sphere_map.html
ijlal99: For `Abandoned Equirectangular`, we have the below license.
https://creativecommons.org/licenses/by-nc-sa/2.0/deed.en
ThreeJS also provides the attribution for this image in a similar way.
https://threejs.org/examples/webgl_panorama_equirectangular.html

For the `Lake Chelan`, we have the below license:
https://unsplash.com/license

paireks (commented):
I like it.
My 2 cents: For me it's not must-have, but I'm thinking if this example would be even better with 1 model loaded inside.

xeolabs:
> I like it. My 2 cents: For me it's not must-have, but I'm thinking if this example would be even better with 1 model loaded inside.

Yeah I agree - @ijlal99 could you please add a model to that example - good for context, but also to ensure all the rendering paths play nicely together

src/viewer/scene/utils/SphereMap.js
```diff
@@ -0,0 +1,60 @@
+import { buildSphereGeometry } from "../geometry/builders/buildSphereGeometry.js";
```
MichalDybizbanskiCreoox, line 1: If you look at the `createSphereMap` function, what it does is it creates a sphere with a material using a texture.
Does this really deserve a dedicated abstraction (as opposed to e.g. one or two elaborate examples)?

And if it did, should this abstraction belong to the `/src/viewer` hierarchy?
It doesn't extend rendering features in any way, it operates on a completely different (higher) level of abstraction.
Given that we really do want to provide such an abstraction, couldn't it be contained inside the `/src/plugins`, `/src/extras`, or even `/examples` hierarchy?
It's worth keeping the SDK short and simple.

MichalDybizbanskiCreoox (changes requested):

src/viewer/scene/utils/SphereMap.js
xeolabs: I agree that it would be better to move the `createSphereMap` function into `/src/extras`.

examples/viewer/sphere_map.html
paireks: As mentioned here: https://github.com/xeokit/xeokit-sdk/pull/1878, probably I wouldn't use content under BY-NC-SA

ijlal99: Removed abandoned equirectangular image as it's not compatible with our license. Thank you!

xeolabs (approved):

PR #1873, opened 2025-04-28
[REFACTOR] Listen only onCameraMatrix

paireks:
This PR is a result of suggestion from Michał to check if we really need to listen all different events or maybe listening to onCameraMatrix is enough. For me it seems that listening to onCameraMatrix is enough, because for:

- mouse interactions
- keyboard interactions
- browser window resizing

it all seems to still trigger the FastNavPlugin mode properly. But please check it before the merge, and let me know if there is some other reason to listen other types that makes it worth keeping them.


xeolabs:
Looks good to me - I can't think a reason to keep the other events in this case.

PR #1877, opened 2025-04-30
XEOK-260 Provide general purpose TransformControl

MichalDybizbanskiCreoox:
This PR introduces a general purpose `TransformControl` abstraction, based on the former `SectionPlanePlugin`'s `Control`.
An example of its use has been added as `examples/annotations/annotations_transformControl.html`.

IMPORTANT:
This implementation changes the `Renderer` class, to run a second "UI" render pass for the `TransformControl` visual components to not be obscured by the scene geometry.
@xeolabs let's discuss this implementation detail, perhaps there's another approach that could be used to achieve the objective.


MichalDybizbanskiCreoox:
Good catch there by @felipedc09!
@xeolabs I think we better incorporate the fix with the commit that introduces a bug, to keep the commit history stable (as in - not merging commits with known errors).

MichalDybizbanskiCreoox:
@xeolabs I've just incorporated @felipedc09's TS fix, also made small adjustments to the examples code, ready to be reviewed and merged.

src/plugins/SectionPlanesPlugin/SectionPlanesPlugin.js
```diff
@@ -234,7 +244,160 @@ class SectionPlanesPlugin extends Plugin {
...
+                                         indices: [0, 1, 2, 2, 3, 0]
+                                     }),
+                                     material: new PhongMaterial(rootNode, {
+                                         emissive: [0, 0.0, 0],
+                                         diffuse: [0, 0, 0],
+                                         backfaces: true
+                                     }),
+                                     opacity: 0.6,
+                                     ghosted: true,
+                                     ghostMaterial: new EmphasisMaterial(rootNode, {
```
xeolabs, line 272: Shouldn't this be xrayed and xrayMaterial?

types/plugins/TransformControl/TransformControl.d.ts
```diff
@@ -0,0 +1,50 @@
+import { Viewer } from "../../viewer";
+
+export declare type TransformControlHandlers = {
+    /** Optional callback for translation change. */
+    setPosition?: (position: number[]) => void;
```
xeolabs, line 5: Can we rename these to the form `onPosition`, `onQuaternion` etc

xeolabs (changes requested):

src/plugins/SectionPlanesPlugin/SectionPlanesPlugin.js
MichalDybizbanskiCreoox: Good call.
Both properties have been carried over from the original `Control` class:
https://github.com/xeokit/xeokit-sdk/blob/ce67c094f3d2e8877c60fc84a272892d59c6e751/src/plugins/SectionPlanesPlugin/Control.js#L383-L391
However it looks like they are not handled in any place, so likely could be removed, right?

xeolabs: Yeah go ahead and delete them if they're not used - looks like dead copy/pasted code, I think from xeogl

xeolabs (approved):
Looks good!

PR #1879, opened 2025-05-09
Turning on the reuse of geometry

paireks:
Hello @xeolabs,

I had to turn it on, because when Diyan  tried these examples with different file, which was quite heavy, then it was crashing browser or showing only part of the model.

My question: is there a scenario when having it turned off may be beneficial?


xeolabs:
Hi @paireks - I can't think of a scenario where turning off would be beneficial. It seems to be redundant here - if the examples still work, then I think we should remove it altogether.

PR #1884, opened 2025-05-21
Allow to read geometry directly from the GPU (VBO triangles layers)

tmarti:
## Context

This is the result of some fantastic conversation with @MathiasTribia and based on the initial code he did in `ArrayBuf.js` but that was not public 😎. Big Kudos Mathias! 🥂 

Now I took the opportunity to work out some details and leave it in a really good working state 🎉. Perfectly working for VBO batching/instancing triangles layers!

## Description

This PR allows to read back geometry data DIRECTLY from the GPU without needing to store all the geometry of loaded models into browser RAM.

Completely avoid the need for `Scene::readableGeometryEnabled`!

It's a tremendous RAM saver in case of needing to read back entities' geometry in order to do further processing/analysis on the Javascript side.

It's very simple to use! Use it like this::

```javascript
const entity = ...;
const { indices, positions } = e.getGeometryData();

// now, have some fun with the geometry in `indices` / `positions` 🎉 
```

The fetch time is independent of the model complexity, and only dependant on the amount of fetched geometry.

Some tests fetching the geometry of around 3k objects indicate the geometry can be fetched for all 3k of them in less than 0.5s! 🚀 

At the moment this is supported by both batching and instancing layers of VBO-Triangles implementations!

What do you think @xeolabs?

### DIFF
--- a/src/viewer/scene/Entity.js
+++ b/src/viewer/scene/Entity.js
@@ -492,6 +492,12 @@ class Entity {
     }
 
     /**
+     * Gets the complete geometry of this entity.
+     */
+    getGeometryData() {
+    }
+
+    /**
      * Destroys this Entity.
      */
     destroy() {
--- a/src/viewer/scene/model/SceneModelEntity.js
+++ b/src/viewer/scene/model/SceneModelEntity.js
@@ -687,6 +687,32 @@ export class SceneModelEntity {
         }
     }
 
+    readGeometryData() {
+        let positionsBase = 0;
+        const positions = [];
+        const indices = [];
+
+        for (let i = 0, len = this.meshes.length; i < len; i++) {
+            const mesh = this.meshes[i];
+
+            if (layer.readGeometryData) {
+                const meshGeom = mesh.layer.readGeometryData(mesh.portionId);
+    
+                for (let j = 0, len = meshGeom.indices.length; j < len; j++) {
+                    indices.push(meshGeom.indices[j]+positionsBase);
+                }
+    
+                for (let j = 0, len = meshGeom.positions.length; j < len; j++) {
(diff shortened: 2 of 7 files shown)


dknezevic:
I really like the option to grab the geometry directly from the GPU as i am currently struggling with the Scene::readableGeometryEnabled option in order to do additional computations on the model.
Can't wait to test it!

tmarti:
Any news here? 😊

xeolabs:
Hi @tmarti, coupld of quick questions, as we work out strategies - 

Would this be doable for DTX as well?
What's your use case for this feature?

thanks

MichalDybizbanskiCreoox:
> I really like the option to grab the geometry directly from the GPU as i am currently struggling with the Scene::readableGeometryEnabled option in order to do additional computations on the model. Can't wait to test it!

Hi @dknezevic, Michał from Creoox here.
What are those issues when using `Scene::readableGeometryEnabled` that you hope this PR would help with?
What are your use cases that'd require accessing geometry data?

tmarti:
@MichalDybizbanskiCreoox, is there something else I can do on my side to make this PR move forward? I'm happy to help if needed. We have some use cases that require fetching geometry from the GPU too to perform analyses based on object meshes at visualization time (backend processing is a no go in these cases to not compilcate the conversion pipeline we have) and are also potentially loading big federated models altogether, which makes `Scene::readableGeometryEnabled` not an option.

Also, I think this PR is very well self-contained and not touching any other existing logic in xeokit-sdk.

Or maybe there is some other good reason from your side, technical of some other, to keep the merge paused at the moment? Would just like to understand it a bit better :)

tmarti:
Thank you so much for merging! 🎉

PR #1921, opened 2025-07-17
fix pivotController position when explorer pane opens and closes

ahhhchuen:
Fix PivotController position when canvas proportion is modified by CSS.
This particularly affect the Xeokit BIM viewer, which implemented this PivotController feature that affected by explorer pane (left side) opening and closing.


xeolabs:
Hi @ahhhchuen , could you please provide a minimal HTML page that embeds the BIMSurfer and reproduces this problem?

These CSS issues arise from various user's unanticipated page layouts, so fixing this for one user could break it for others, so this would be helpful to see if it's possible to include this PR.

thanks

ahhhchuen:
by the way, do you have people caring xeokit BIM viewer repo? I have a PR there without response yet.

PR #1922, opened 2025-07-17
Related to XCD-333 <- refreshing the camera on the adding and removing of section planes

paireks:
Hey @xeolabs & @MichalDybizbanskiCreoox,

this PR fixes the XCD-333, this is the behaviour before the change:
2025-07-17_19h59_10

And this is after:
2025-07-17_19h31_59

In other words what it does is it tries to force to redraw the measurements when you remove the section plane, not only when you move the camera.

It also should work once when you add the section plane directly where the measurement are. Right now you also have to move the camera first to see the effect.

Having said that I'm not sure if this is an elegant way of solving this. I can't see any other place where Scene calls camera to be updated, so maybe there is something better? If yes - then please let me know. Will switch from draft after review from you both.

Thank you for help!

### DIFF
--- a/examples/slicing/SectionPlanesPlugin_withMeasurement.html
+++ b/examples/slicing/SectionPlanesPlugin_withMeasurement.html
@@ -1,0 +1,201 @@
+<!doctype html>
+<html>
+<head>
+    <meta charset="utf-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1">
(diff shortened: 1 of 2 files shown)


MichalDybizbanskiCreoox (changes requested):
It'd be more consistent with existing logic to rely on the `sectionPlaneCreated` and `sectionPlaneDestroyed` inside the `Dot3D`, `Label3D`, and `Wire3D` abstractions, as they already rely on the `sectionPlaneUpdated` events.

PR #1929, opened 2025-08-01
fix: Update links in README and index.html for consistency; change im…

damianbobrowski:
…age source in getSnapshot example


examples/snapshots/getSnapshot_png_withAnnotations.html
```diff
@@ -266,7 +266,7 @@ <h3>Resources</h3>
                 title: "The West wall",
                 description: "Annotations can contain<br>custom HTML like this<br>image:",
                 markerBGColor: "red",
-                imageSrc: "https://xeokit.io/img/docs/BIMServerLoaderPlugin/schependomlaan.png"
+                imageSrc: "https://old.xeokit.io/img/docs/BIMServerLoaderPlugin/schependomlaan.png"
```
MichalDybizbanskiCreoox, line 269: It's likely that `old.xeokit.io` will be turned off eventually, and the image will become missing again.
What would it take to have the `schependomlaan.png` image hosted by the `xeokit.io` domain?
Since the page that is tracked by the repo (`getSnapshot_png_withAnnotations.html`) refers to the image, perhaps the image could be placed alongside in the same directory and be tracked as well?

MichalDybizbanskiCreoox (changes requested):

damianbobrowski:
@MichalDybizbanskiCreoox 
PR contain one commit and its rebased with master

PR #1930, opened 2025-08-05
refactor: update type definitions for touchPointSelector to be generic

damianbobrowski:
@MichalDybizbanskiCreoox 
I got hint, which initially comes from you that:

> the type returned by `Ray2WorldPos` should be arbitrary but compatible with the type passed as the second parameter to `onChange` and `onCommit`

that is what I tried to introduce in type declarations.

****
Shall we narrow down generic type `<T>` to `<T extends Vec3 | PickResult>`, based on available examples:
- `Vec3` - `annotations_createWithTouch.html`
- `PickResult` - `SectionPlanesPlugin_createWithTouch.html`

or left in more generic?

### DIFF
--- a/types/plugins/lib/ui/index.d.ts
+++ b/types/plugins/lib/ui/index.d.ts
@@ -1,12 +1,15 @@
 import { Viewer } from "../../../viewer";
 import { PointerCircle } from "../../../extras/PointerCircle";
 
+type Vec2 = number[]
+type Vec3 = number[]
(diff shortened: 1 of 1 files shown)


MichalDybizbanskiCreoox:
> Shall we narrow down generic type `<T>` to `<T extends Vec3 | PickResult>`, based on available examples or left in more generic?
There's no dependency on the specific types inside the Ray2WorldPos or around it, so I think more generic makes sense.

PR #1932, opened 2025-08-07
fix(XEOK-100): update build scripts for improved command chaining and…

damianbobrowski:
… cleanup

- Changed the build script to use '&&' instead of ';' for for cross-platform compatibility and better error handling.
- Replaced 'rm -Rf' with 'rimraf' for cross-platform compatibility in docs cleanup.
- Updated the changelog script to use '&&' for improved command chaining.
- Added 'rimraf' as a dependency to the project.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@xeokit/xeokit-sdk",
-  "version": "2.6.82",
+  "version": "2.6.87",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@xeokit/xeokit-sdk",
-      "version": "2.6.82",
+      "version": "2.6.87",
       "license": "AGPL-3.0",
       "dependencies": {
         "@creooxag/cxconverter": "^0.0.3-alpha",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -68,13 +69,13 @@
...
-    "esdoc-ecmascript-proposal-plugin": "1.0.0",
-    "esdoc-publish-html-plugin": "1.1.2",
-    "esdoc-standard-plugin": "1.0.0",
+    "esdoc": "^1.1.0",
+    "esdoc-custom-theme": "^1.4.2",
+    "esdoc-publish-html-plugin": "^1.1.2",
+    "esdoc-standard-plugin": "^1.0.0",
     "eslint": "^8.13.0",
     "parse5": "^7.0.0",
+    "rimraf": "^6.0.1",
```
MichalDybizbanskiCreoox, line 78: What cross-platform compatibility issues are we addressing by adding this dependency?

package-lock.json
```diff
@@ -1,18 +1,19 @@
 {
   "name": "@xeokit/xeokit-sdk",
-  "version": "2.6.105",
+  "version": "2.6.87",
```
MichalDybizbanskiCreoox, line 3: Do these ~900 new lines result from adding a single dependency?
~The versions are also very outdated, should this file be even included in the commit?~ Edit: Just noticed this PR is from past year, and the target branch is `master-old`.

package.json
drashevski: just in case mentioning @damianbobrowski

damianbobrowski:
this PR is not relevant any more as most of this scripts was removed.
closing this PR

PR #1937, opened 2025-08-19
fix(XCD-350): update clipSegment logic to improve segment clipping conditions

damianbobrowski:
When  viewer.camera.projection = "ortho", then for "Y" measurement `denom` = 0. 
Its edge case which makes, disappearing Y `Wire` and `label`.

 `(denom => 0)` seams to work, but maybe we can compare direct values which are used to compute values in next step.


damianbobrowski:
Finally I move logic of calculating tmin, tmax into separated function `liangBarskyClip`. Its body explicitly shows, steps. Hope it is easier to follow, and understand geometrically what it does.

PR #1955, opened 2025-10-01
SectionCaps changes: XEOK-361, minor fixes and refactoring

MichalDybizbanskiCreoox:
This PR :
1. Implements [XEOK-361] Improve feedback (throw exception) when SectionCaps used with readableGeometryEnabled: false,
2. Fixes minor SectionCaps issues,
3. Refactors SectionCaps to isolate individual entity processing (as opposed to combining all entities in multiple stages), so that the cap generation is simpler to evaluate and better adapted to a future migration to webworkers.

!!! IMPORTANT
The first point (XEOK-361) is a breaking change.
Code that previously would silently fail will now throw an error.
This should make the readableGeometryEnabled requirement more clear.


MichalDybizbanskiCreoox:
Deprecated by #1979

PR #1964, opened 2025-10-07
[FIX] SceneModelEntity primitive count broken (NaN) 

Kurtil:
Use [`mesh.numTriangles`](https://github.com/xeokit/xeokit-sdk/blob/master/src/viewer/scene/model/SceneModelMesh.js) instead of unknown `mesh.numPrimitives` to compute `sceneModelEntity.numPrimitives`. Maybe it works in some scenario if `numPrimitives` was added by parser or something else, but in my case it was broken loading LAZ (point cloud) models. Indeed, as `sceneModelMesh` is `undefined`, the computed `numPrimitives` is `NaN` .

### DIFF
--- a/src/viewer/scene/model/SceneModelEntity.js
+++ b/src/viewer/scene/model/SceneModelEntity.js
@@ -55,7 +55,7 @@ export class SceneModelEntity {
             const mesh = this.meshes[i];
             mesh.parent = this;
             mesh.entity = this;
-            this._numPrimitives += mesh.numPrimitives;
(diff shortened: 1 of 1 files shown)


MichalDybizbanskiCreoox:
Hi @Kurtil,
Thank you for this PR.
Just wanted to clarify - you load the LAZ model using xeokit's `LASLoaderPlugin`, right?
We'll have to fix the core problem here, that is of `mesh.numPrimitives` being (wrongly) `undefined`.
Best,
  Michał

Issue #1949, opened 2025-09-22
Memory leak issue after viewer destruction


intrida-dev:
We use create and destroy Xeokit viewers as per user actions. We observed that after destroying viewer, memory is not getting release. Few object references are still persisted.

Refer the SS as below:

<img width="507" height="207" alt="Image" src="https://github.com/user-attachments/assets/23d9aa79-e7f1-4e06-879d-8481b855e379" />

Before loading the model, 20.4MB memory was consumed. And after model is loaded and destroyed using _viewer.destroy();_, it is consuming 414MB. This becomes serious issue for big datasets.

To reproduce this issue, I have attached the sample code. Please follow following steps to reproduce:

1. Use following commands to run the code:
npm install;
npm start;
2. Capture a Heap snapshot from Memory tab.
3. Now create Xeokit viewer and load model by clicking on 'Load' button.
4. Once model is loaded, click on 'Destroy' button to destroy the viewer.
5. Capture another Heap snapshot from Memory tab.
6. Compare these 2 and see huge heap memory increase in second.

Please let us know, how can we better manage this.
**Note: I couldn't upload the model which I have been using for testing this, because of the size issue.**

[xeokit-memory.zip](https://github.com/user-attachments/files/22466193/xeokit-memory.zip)

PR #1989, opened 2025-11-26
#1949 SceneModelMesh's layer reference removal

paireks:
Hey,

after memory trouble shooting I've noticed big difference only after this layer reference removal from SceneModelMesh so far.

For dtx WestRiverSideHospital heap snapshot after viewer.destroy():
- before this change : 905 MB
- after this change : 92 MB

For federated Clinic model heap snapthot after viewer.destroy():
- before this change: 143 MB
- after this change: 29 MB

### DIFF
--- a/src/viewer/scene/model/SceneModelMesh.js
+++ b/src/viewer/scene/model/SceneModelMesh.js
@@ -415,5 +415,7 @@ export class SceneModelMesh {
      */
     _destroy() {
         this.model.scene._renderer.putPickID(this.pickId);
+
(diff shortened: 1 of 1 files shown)


xeolabs:
Thanks @paireks - so a reference was holding on to memory by the looks of it. In theory this should have been solved by  SceneModel.destroy nulling all references to/amongs the components, but it seems something was holding onto a reference. There seems to be a sticky reference somewhere among those components, but this PR certainly doesn't hurt.

PR #2004, opened 2025-12-15
XCD-415 Add more descriptive explanation of the Curve's length getter

paireks:


src/viewer/scene/paths/Curve.js
```diff
@@ -52,9 +52,15 @@ class Curve extends Component {
...
     /**
-     * Gets the length of this Curve.
+     * Gets the approximate length of this Curve.
      *
-     * @returns {Number} The Curve length.
+     * This method calculates the length by discretizing Curve into hardcoded number of straight segments. Then it sums
+     * lengths of these straight segments.
+     *
+     * If you need more control over the precision of the length calculation, we strongly encourage you to build
+     * your own custom function.
```
MichalDybizbanskiCreoox: Maybe include an example here?
E.g.
```js
function getSplineLength(spline, divisions) { // higher `divisions` gives better accuracy
    let p0 = spline.getPoint(0);
    let sum = 0;
    for (let i = 1; i <= divisions; i++) {
        const p1 = spline.getPoint(i / divisions);
        sum += math.distVec3(p0, p1);
        p0 = p1;
    }
    return sum;
}
```

MichalDybizbanskiCreoox (approved):

PR #2035, opened 2026-04-08
Support GLTFLoaderPlugin::load::loadersGl parameter to load alternative loaders.gl version

MichalDybizbanskiCreoox:
The PR makes it possible for users to inject their own loaders.gl library of any version to be used by `GLTFLoaderPlugin`.
The `GLTFLoaderPlugin::load::loadersGl` can be passed as a `{ core, gltf }` object that contains loader.gl’s core and gltf modules respectively.
If `loadersGl` object is provided, the `GLTFLoaderPlugin` will use it instead of the default v4.3.4 version.

### DIFF
--- a/examples/buildings/glb_alternative_loadersGl.html
+++ b/examples/buildings/glb_alternative_loadersGl.html
@@ -1,0 +1,78 @@
+<!doctype html>
+<html>
+<head>
+    <meta charset="utf-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1">
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <title>xeokit Example</title>
+    <link href="../css/pageStyle.css" rel="stylesheet"/>
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/5.13.0/js/all.min.js"></script>
+</head>
+<body>
+<input type="checkbox" id="info-button"/>
+<label for="info-button" class="info-button"><i class="far fa-3x fa-question-circle"></i></label>
+<canvas id="myCanvas"></canvas>
+<div class="slideout-sidebar">
+    <img class="info-icon" src="../../assets/images/gltf_logo.png"/>
+    <h1>GLTFLoaderPlugin</h1>
+    <h2>Using an alternative loaders.gl version</h2>
+    <p>In this example, we're using a <a
+            href="../../docs/class/src/plugins/GLTFLoaderPlugin/GLTFLoaderPlugin.js~GLTFLoaderPlugin.html"
+            target="_other">GLTFLoaderPlugin</a> to load the <a
+            href="https://sketchfab.com/bimcc"
+            target="_other">House Plan</a> demo model from
+        binary glTF (.glb).</p>
+    <h3>Components used</h3>
+    <ul>
+        <li>
+            <a href="../../docs/class/src/viewer/Viewer.js~Viewer.html"
+               target="_other">Viewer</a>
+        </li>
+        <li>
+            <a href="../../docs/class/src/plugins/GLTFLoaderPlugin/GLTFLoaderPlugin.js~GLTFLoaderPlugin.html"
+               target="_other">GLTFLoaderPlugin</a>
+        </li>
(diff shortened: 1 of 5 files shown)


examples/buildings/glb_alternative_loadersGl.html
```diff
@@ -0,0 +1,78 @@
...
+
+    // User-provided loaders.gl
+    import * as loadersGlCore from "https://cdn.jsdelivr.net/npm/@loaders.gl/core@4.4.1/+esm";
+    import * as loadersGlGltf from "https://cdn.jsdelivr.net/npm/@loaders.gl/gltf@4.4.1/+esm";
+
+    // Alternatively e.g.
+    // import * as loadersGlCore from "https://esm.sh/@loaders.gl/core@4.4.1";
+    // import * as loadersGlGltf from "https://esm.sh/@loaders.gl/gltf@4.4.1";
+
+    import {Viewer, GLTFLoaderPlugin} from "../../src/index.js";
```
paireks: Remember to change here the path to dist

src/plugins/GLTFLoaderPlugin/GLTFLoaderPlugin.js
```diff
@@ -323,6 +323,7 @@ class GLTFLoaderPlugin extends Plugin {
      * @param {Boolean} [params.globalizeObjectIds=false] Indicates whether to globalize each {@link Entity#id} and {@link MetaObject#id}, in case you need to prevent ID clashes with other models.
      * @param {*} [params.parseOptions={}] Options to pass to loaders.gl parse method, eg. ````{ gltf: { excludeExtensions: { "KHR_texture_transform": false } } }````.
      * @param {Boolean} [params.entityPerMesh=false] Create an entity for each mesh, instead of grouping leaf meshes under their common entity.
+     * @param {*} [params.loadersGl={core,gltf}] Alternative user-provided loaders.gl library.
```
paireks, line 326: To decide what to do when someone inputs here the version not compatible with the one we currently support. Or once we will upgrade the version internally, but someone still loads the alternative version which was compatible at first, but would no longer be compatible after an upgrade.

examples/buildings/glb_alternative_loadersGl.html
MichalDybizbanskiCreoox: Good catch, thanks

src/plugins/GLTFLoaderPlugin/GLTFLoaderPlugin.js
MichalDybizbanskiCreoox: I think we have to hope for loader.gl's interface not to change, and for the users to check compatibility at their discretion.
We plan to implement at some point an alternative gltf loader that doesn't depend on loaders.gl.

paireks: Looking at these logs: https://loaders.gl/docs/upgrade-guide I'm not sure if we can hope for lack of breaking changes.

Problem is such additional ability (ability to load alternative loaders.gl version) for me it suggests you can plug any version there and it will work (any existing version from the past and any version which will appear in the future), but is it true?

MichalDybizbanskiCreoox: I don't think there is any practical way to reliably manage compatibility tests of custom loaders.gl versions passed to the GLTFLoaderPlugin.
We'd either end up excluding some valid ones, and/or accepting invalid ones.
Users are free to pass an arbitrary version, but the onus is on them to make sure it's a valid one.
The `loadersGl` parameter is not a feature required to use - after all we still provide a fallback to the hardcoded loaders.gl version.
Even if we didn't provide a fallback - which me might end up doing in the future - the best we could do would be to name a version (or a number of versions) that we had verified to be cooperative with the plugin.
Preventing users from using a version which we didn't verify, but which could be good enough for the plugin, would be overprotective IMO.

MichalDybizbanskiCreoox: There has just been reported another issue asking for this feature.

PR #2063, opened 2026-06-01
XEOK-465 Extend reuseGeometries to function type

MichalDybizbanskiCreoox:
Accept a function as the `XKTLoaderPlugin::load::reuseGeometries` parameter, to allow expressing a conditional strategy for whether a geometry should be instanced or not.


xeolabs (approved):
👍👍

PR #2072, opened 2026-07-15
Create publiccode.yml for xeokit SDK

drashevski:
Add publiccode.yml for xeokit SDK with detailed metadata.

### DIFF
--- a/publiccode.yml
+++ b/publiccode.yml
@@ -1,0 +1,99 @@
+publiccodeYmlVersion: "0.7"
+
+name: xeokit SDK
+
+url: https://github.com/xeokit/xeokit-sdk
+landingURL: https://xeokit.io/
+
+softwareVersion: "2.6.112"
+releaseDate: "2026-06-25"
+developmentStatus: stable
+softwareType: library
+
+platforms:
+  - web
+
+categories:
+  - application-development
+  - web-development
+  - visualization
+  - cad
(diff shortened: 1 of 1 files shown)


publiccode.yml
```diff
@@ -0,0 +1,101 @@
+publiccodeYmlVersion: "0.7"
+
+name: xeokit SDK
+
+url: https://github.com/xeokit/xeokit-sdk
+landingURL: https://xeokit.io/
+
+softwareVersion: "2.6.112"
```
paireks, line 8: This and releaseDate is a bit problematic if we would have to update it every time. @mmilian - is there a way to update it automatically? Or maybe it's not necessary to update it every time?

paireks (commented):
Detail which Michał has taught me: if possible it could be squashed into a single commit to not pollute the git history.

publiccode.yml
drashevski: so, when merging, just select "squash and merge". About the release date - I don't mind whatever option you pick.

mmilian: we could prepare a githook script, which notify us about this file must be updated whenever package.json version is updated (it could potentially even update this file automatically)

MichalDybizbanskiCreoox: Could the `publiccode.yml` be edited by the Release GH action?

